PR #66, opened 2017-07-13
Merge capapi history

anastasia:
This is everything from capapi, it's currently running on port 8080 alongside capstone. 
I'll start merging individual files to capstone, and have smaller pull requests from here on out.
This is a very big commit — having the preserved history would be nice, but if it's too much we can also look at squashing them and keep capapi as an archived repo.


.gitignore
```diff
@@ -3,9 +3,15 @@ settings.py
 __pycache__
 .vagrant/
 .virtualenvs/
+*.pyc
 
 db.sqlite3
 
-# pytest
```
ChefAndy, line 9: Put back

capstone/cap/models.py
```diff
@@ -266,6 +270,13 @@ class CaseMetadata(models.Model):
     def __str__(self):
         return self.slug
 
+class Citation(models.Model):
+    type = models.CharField(max_length=100)
```
ChefAndy, line 274: Make choice!

PR #67, opened 2017-07-13
update - add citation model, update fields

anastasia:


capstone/cap/models.py
```diff
@@ -251,22 +251,36 @@ class CaseMetadata(models.Model):
     slug = models.SlugField(unique=True, max_length=255)
     first_page = models.IntegerField(null=True, blank=True)
     last_page = models.IntegerField(null=True, blank=True)
-    jurisdiction = models.ForeignKey('Jurisdiction', null=True, related_name='%(class)s_jurisdiction', on_delete=models.SET_NULL)
-    citation = models.CharField(max_length=255, blank=True)
+    jurisdiction_xml = models.CharField(max_length=255, blank=True)
```
ChefAndy: I think that we should convert the xml to json, and then have this be a json or jsonb field. If that doesn't work for some reason, I think we should at least use the Postgres XML field type so we can query the enclosed data directly rather than having to grab the string and parse it locally. Thoughts?

ChefAndy:
Quick question on the jurisdiction_xml field format.

capstone/cap/models.py
anastasia: Sorry, I thought this would be just a string field that we would remove later on, the `_xml` just denoting that this is what we got from the xml directly, before we are able to locate the actual jurisdiction model this case belongs to and set it as foreign key. Maybe the naming is confusing and it should be something like `jurisdiction_string`? Is there a reason to treat the `jurisdiction_xml` field as anything but temporary?

ChefAndy: Oh, I see. That makes sense. I don't think we should label it XML unless there's XML in there. Maybe jursdiction_value?

Issue #64, opened 2017-07-11
Change citation in CaseXML model 


ChefAndy:
In the CaseXML model, we want to split up citation into two fields:
1) a plain text field for the category="official" citation
2) a json field with all citations and their related metadata

PR #69, opened 2017-07-31
CAP API

anastasia:
This pull request adds the capapi app.
Other things of note:
- capapi has its own database of the same name, make sure to `psql -c "CREATE DATABASE capapi;"`
- I've renamed the cap app to capdb for clarity
- Added the Jurisdiction model in capdb
- Added APIUser and APIUser manager models in capapi
- All API related keys that live in settings are namespaced API_{key}
- Added `fab migrate` command to deal with all the different databases and apps
- Added process_metadata.py in scripts to parse case xml and return all the data we need for the CaseMetadata objects
- Added many to many `jurisdictions` field in Reporter model

finishes #68, finishes #64


capstone/scripts/ingest_tt_data.py
```diff
@@ -2,19 +2,19 @@
...
+from capdb.models import VolumeMetadata, TrackingToolUser, Reporter, ProcessStep, TrackingToolLog, BookRequest
 from tracking_tool.models import BookRequests, Eventloggers, Hollis, Pstep, Reporters, Users, Volumes
 
 
 user_field_map = {'id': 'id', 'privlevel': 'privilege_level', 'email': 'email', 'created_at': 'created_at', 'updated_at': 'updated_at' }
 book_request_field_map = {'id': 'id', 'updated_by': 'updated_by', 'created_at': 'created_at', 'updated_at': 'updated_at', 'recipients': 'recipients', 'from_field': 'from_field', 'mail_body': 'mail_body', 'note': 'note', 'send_date': 'send_date', 'label': 'label', 'sent_at': 'sent_at', 'subject': 'subject', 'delivery_date': 'delivery_date'} 
 pstep_field_map = {'step_id': 'step', 'name': 'label', 'prereq': 'prerequisites', 'desc': 'description', 'created_at': 'created_at', 'updated_at': 'updated_at'}
-reporter_field_map= {'id': 'id', 'state': 'jurisdiction', 'reporter': 'full_name', 'short': 'short_name', 'start_date': 'start_year', 'end_date': 'end_year', 'volumes': 'volume_count', 'created_at': 'created_at', 'updated_at': 'updated_at', 'notes': 'notes'}
```
anastasia, line 12: removed from mapping state to jurisdiction because reporter now has ManyToMany Jurisdiction field, which needs to be set like this: reporter_instance.jurisdictions.add(new_jurisdiction)

anastasia:
steps to merge:

-- migrate backwards 
-- psql rename tables, cap_{ name } to capdb_{ name }
-- merge pull request
-- migrate forwards

capstone/capapi/admin.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class APIUserAdmin(admin.ModelAdmin):
+    readonly_fields = ('date_joined', 'api_key')
+    list_display = ('email', 'last_name', 'first_name', 'api_key')
+
+    def api_key(self, instance):
+        return instance.get_api_key()
+    api_key.short_description = "API Key"
+    fields = (
```
jcushman: Nitpick: it's a little confusing to have `fields` down below the functions -- typically would go up with the other attributes at the top of the class.

capstone/capapi/case_views.py
```diff
@@ -0,0 +1,90 @@
...
+from django_filters.rest_framework import DjangoFilterBackend
+from rest_framework import renderers, viewsets, mixins, filters as rs_filters
+
+from . import permissions, resources, serializers, filters
+from capdb import models
+
+logger = logging.getLogger(__name__)
+
+
+class JSONResponse(HttpResponse):
```
jcushman: Is this redundant of https://docs.djangoproject.com/en/1.11/ref/request-response/#jsonresponse-objects ?

If it's doing something different, maybe add a comment explaining.

capstone/capapi/case_views.py
```diff
@@ -0,0 +1,90 @@
...
+    ordering = ('decisiondate',)
+
+    def list(self, *args, **kwargs):
+        if not self.request.query_params.get('type') == 'download':
+            return super(CaseViewSet, self).list(*args, **kwargs)
+        else:
+            return self.download_many()
+
+    def retrieve(self, *args, **kwargs):
+        if self.request.query_params.get('type') and self.request.query_params.get('type') == 'download':
```
jcushman: Nitpick: simplify this to `if self.request.query_params.get('type') == 'download':` ?

capstone/capapi/case_views.py
```diff
@@ -0,0 +1,90 @@
...
+            return self.download_one(**kwargs)
+        else:
+            return super(CaseViewSet, self).retrieve(*args, **kwargs)
+
+    def download_cases(self, caseids_list, blacklisted_case_count):
+        try:
```
jcushman: Is this try-except helping? I'm not sure what it's for.

capstone/capapi/routers.py
```diff
@@ -0,0 +1,44 @@
...
+        """Determine if relationship is allowed between two objects."""
+        # Allow any relation between two models that are both in the capapi app.
+        if obj1._meta.app_label == 'capapi' and obj2._meta.app_label == 'capapi':
+            return True
+        # No opinion if neither object is in the capapi app (defer to default or other routers).
+        elif 'capapi' not in [obj1._meta.app_label, obj2._meta.app_label]:
+            return None
+
+        # Block relationship if one object is in the capapi app and the other isn't.
+            return False
```
jcushman: Extra indent breaks the logic here.

(Also style nitpick: maybe reverse the elif test so all of these functions end with `return None`?)

[Edit: below I suggest abstracting out this whole class, so skip this style nitpick ...]

capstone/capapi/routers.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def allow_migrate(self, db, app_label, model_name=None, **hints):
+        """Ensure that the capapi app's models get created on the right database."""
+        if db == 'capapi' and app_label == 'capapi':
+            return True
+        if app_label == 'capapi':
```
jcushman: This last if-elif is a little confusing as a way to express this test -- I think what it boils down to is return false if db or app_label is capapi, which is the same test you're running up in `allow_relation` but expressed in a very different way.

Suggestion: this router ultimately implements exactly the same logic as `tracking_tool.routers.TrackingToolDatabaseRouter`, except with the string 'capapi' instead of 'tracking_tool'. Rather than have two redundant routers, could we create a `class SeparateDatabaseRouter` that implements the TrackingToolDatabaseRouter logic with `self.database_name` in place of `'tracking_tool'`, and subclass it as `class CAPAPIRouter(SeparateDatabaseRouter): database_name = 'capapi'`, `class TrackingToolDatabaseRouter(SeparateDatabaseRouter): database_name = 'tracking_tool'`?

capstone/capapi/serializers.py
```diff
@@ -0,0 +1,144 @@
...
+        fields = ('email', 'password')
+        write_only_fields = ('password')
+        lookup_field = 'email'
+
+    def verify_with_password(self, email, password):
+        user = APIUser.objects.get(email=email)
+        correct_password = user.check_password(password)
+        if not correct_password:
+            raise serializers.ValidationError('Invalid password')
+        return user
```
jcushman, line 144: (Github doesn't let me leave comments on the actual files, so this applies to the next files down below.) It looks like this pull req includes a bunch of files in `.sass-cache` that should be .gitignored? Or maybe I'm not understanding the render pipeline here.

capstone/capapi/views.py
```diff
@@ -0,0 +1,121 @@
...
+from rest_framework.permissions import AllowAny
+from rest_framework.parsers import JSONParser, FormParser
+
+from . import models, serializers, resources, permissions
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+class JSONResponse(HttpResponse):
```
jcushman: Same question as above -- redundant of Django version?

capstone/capapi/views.py
```diff
@@ -0,0 +1,121 @@
...
+                    'password': request.data.get('password'),
+                    'first_name': request.data.get('first_name'),
+                    'last_name': request.data.get('last_name')
+                })
+                content = {
+                    'status': 'Success!',
+                    'message': 'Thank you. Please check your email %s for a verification link.' % user.email
+                }
+                return Response(content, template_name='sign-up-success.html')
+            except IntegrityError as e:
```
jcushman: Is this the flow for a duplicate email address? Does it make sense for it to use a different flow/template from other validation errors? If so maybe add a comment to explain ...

jcushman (changes requested):
What a lot of excellent work! I threw in some comments and suggestions for the part I've reviewed so far, before the baby freaked out. :) Will try to get to the rest this evening.

capstone/capapi/views.py
```diff
@@ -0,0 +1,121 @@
...
+        serializer = serializers.LoginSerializer(data=request.data)
+        if serializer.is_valid():
+            try:
+                user = serializer.verify_with_password(email=request.data.get('email'), password=request.data.get('password'))
+
+                api_key = user.get_api_key()
+                if api_key:
+                    # update case allowance before sending back
+                    user.update_case_allowance()
+                    user = models.APIUser.objects.get(email=user.email)
```
jcushman: Nitpick: is this the same as `user.refresh_from_db()`? If so that seems clearer.

capstone/capapi/wsgi.py
```diff
@@ -0,0 +1,9 @@
+import os
```
jcushman: I think this is a redundant file -- we already have a wsgi.py, and whitenoise doesn't have to be added to wsgi.py anymore IIRC.

capstone/capdb/models.py
```diff
@@ -222,36 +235,68 @@ class Court(models.Model):
...
+        try:
+            case_metadata.jurisdiction = Jurisdiction.objects.get(name=data["jurisdiction"])
+        except ObjectDoesNotExist:
+            pass
+
+        court_name = data["court"]["name"]
+        court_name_abbreviation = data["court"]["name_abbreviation"]
+
+        if court_name:
+            court, created = Court.objects.get_or_create(name=court_name)
```
jcushman: Should this include `(name=court_name, jurisdiction=case_metadata.jurisdiction)`? Same for the get_or_create below.

capstone/capdb/models.py
```diff
@@ -268,14 +313,20 @@ class CaseMetadata(models.Model):
...
-    cite = models.CharField(max_length=255, unique=True)
-    slug = models.SlugField(unique=True, max_length=255)
+    cite = models.CharField(max_length=255)
+    slug = models.SlugField(max_length=255, unique=True)
+
+    @classmethod
+    def create(cls, cite, citation_type):
+        citation = Citation(cite=cite, type=citation_type)
+        citation.slug = generate_unique_slug(Citation, 'slug', cite)
+        return citation
```
jcushman: I don't think this approach works for setting the slug field -- it requires all callers to then use the nonstandard `Citation.create()` method to create Citations. In particular I think it doesn't set the slug for `citation, created = Citation.objects.get_or_create(cite=data["citation"], type=["citation_type"])` up above. I'd prefer something like

```
    def save(self, *args, **kwargs):
        if not self.id and not self.slug:
            self.slug = generate_unique_slug(Citation, 'slug', self.cite)
        super(Citation, self).save(*args, **kwargs)
```

capstone/config/settings/settings_base.py
```diff
@@ -121,9 +139,36 @@
...
+### CAP API settings ###
+
+API_CASE_DAILY_ALLOWANCE = 500
+API_CASE_EXPIRE_HOURS = 24
+API_BASE_URL = 'http://localhost:8000'
+API_BASE_URL_ROUTE = '/api'
+API_VERSION = 'v1'
+
+API_FULL_URL = os.path.join(API_BASE_URL_ROUTE, API_VERSION)
+ADMINS = [('Caselaw Access Project', 'info@capapi.org')]
```
jcushman: Maybe move this, it's a little confusing to have it among the API_ settings

capstone/fabfile.py
```diff
@@ -50,17 +54,31 @@ def init_db():
     local("python manage.py migrate")
```
jcushman: Can now replace these three lines with `migrate()`

capstone/scripts/process_metadata.py
```diff
@@ -0,0 +1,58 @@
+from datetime import datetime
+from pyquery import PyQuery as pq
+
+
+namespaces = {
```
jcushman: These can be imported from scripts.helpers

jcushman (commented):
More nitpicks ...

capstone/capapi/case_views.py
anastasia: I'm going to delete the download functionality for now — I meant to do that in this pull request but looks like I missed some code. I'm going to write another pull request (a much smaller one in scope this time) for _just_ the downloading of cases.

capstone/capapi/case_views.py
anastasia: yes! removing download functionality for now (see below)

capstone/capapi/routers.py
anastasia: ooooh that's really neat. I'll do that.

anastasia: can you think of a natural home for the class SeparateDatabaseRouter? scripts doesn't feel right. Maybe config/routers? Or maybe we create a separate directory for this?

capstone/capapi/views.py
anastasia: aha, yes! was not aware of this method.

capstone/capdb/models.py
anastasia: yes! oops

anastasia: actually I'm currently setting jurisdiction to case_metadata only if it's found above:
```
        try:
            case_metadata.jurisdiction = Jurisdiction.objects.get(name=data["jurisdiction"])
        except ObjectDoesNotExist:
            pass
```
Maybe it's safer to set the jurisdiction onto court if created, and if jurisdiction explicitly exists

jcushman: So the thing is, if two courts have the same name but different jurisdiction, are they the same court? I'm guessing no. So I think jurisdiction wants to be in the `get_or_create()`?

Off the top of my head, I think it's fine to pass in `jurisdiction=case_metadata.jurisdiction` even if `case_metadata.jurisdiction` is `None` -- I don't think you need a special case for that.

BTW a nice little shortcut with this "might exist" stuff is to use `first` instead of `get`:

`case_metadata.jurisdiction = Jurisdiction.objects.filter(name=data["jurisdiction"]).first()`

As long as `name` is unique, that has the same effect of either being the thing if it exists or `None`, but without the try-except.

capstone/capapi/routers.py
jcushman: config/ sounds good to me! At least one person on stackoverflow thought so: https://stackoverflow.com/a/34640012/307769

jcushman (approved):

PR #85, opened 2017-09-28
Refactor slug creation

jcushman:
There's an issue with `update_case_metadata` where it always creates a new CaseMetadata instead of updating the existing one. This is because it finds CaseMetadata based on a newly-generated slug, which is required to be unique, so it never finds the existing one.

Changes here:

* `update_case_metadata` finds the existing CaseMetadata, by not running get_or_create based on slug.
* `generate_unique_slug` accepts an instance instead of a model, so it can ignore database entries with the same ID.
* when setting CaseMetadata.citations, `update_case_metadata` replaces existing citations instead of adding.
* Add test to try modifying the xml of a CaseXML and rerunning update_case_metadata.

### DIFF
--- a/capstone/capdb/models.py
+++ b/capstone/capdb/models.py
@@ -1,9 +1,8 @@
 from django.contrib.postgres.fields import JSONField, ArrayField
 from django.db import models
-from django.core.exceptions import ObjectDoesNotExist
 
 from scripts.process_metadata import get_case_metadata
-from .utils import generate_unique_slug, get_citation_to_slugify
+from .utils import generate_unique_slug
 from scripts.helpers import *
 
 ### helpers ###
@@ -246,10 +245,7 @@ class Court(models.Model):
 
     def save(self, *args, **kwargs):
         if not self.id and not self.slug:
-            if self.name_abbreviation:
-                self.slug = generate_unique_slug(Court, 'slug', self.name_abbreviation)
-            else:
-                self.slug = generate_unique_slug(Court, 'slug', self.name)
+            self.slug = generate_unique_slug(self, self.name_abbreviation or self.name)
         super(Court, self).save(*args, **kwargs)
 
     def __str__(self):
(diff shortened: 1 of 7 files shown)


capstone/capdb/models.py
```diff
@@ -307,11 +302,17 @@ def update_case_metadata(self):
         case_metadata.name = data["name"]
         case_metadata.name_abbreviation = data["name_abbreviation"]
 
-        for citation in citations:
-            case_metadata.citations.add(citation)
+        # set slug to official citation, or first citation if there is no official
+        citation_to_slugify = next((c for c in citations if c.type == 'official'), citations[0])
```
anastasia, line 306: Sweet!! I didn't know about this method!

capstone/capdb/utils.py
```diff
@@ -1,36 +1,30 @@
...
+def generate_unique_slug(instance, raw_string, field='slug', max_tries=1000):
     """
-    :param field: the field requiring unique slug
-    :param raw_string: basis for slug
-    :return:
-    Rejection sampling: generate unique slug, check for uniqueness
-    if found in model, resample
+        Get a unique slug for instance by sluggifying raw_string, checking database, and appending -1, -2 etc. if necessary.
+        When checking uniqueness, ignore this object itself if it already exists in the database.
+
```
anastasia, line 8: Is doing the pycharm recommended :param param_name: thing a faux pas?

capstone/capdb/utils.py
```diff
@@ -1,36 +1,30 @@
...
-        return slug
-    else:
-        count += 1
-        return generate_unique_slug(model, field, raw_string, count=count)
+        # ORM query for objects of the same model, with the same slug
+        found = type(instance).objects.filter(**{field: slug})
 
+        # Exclude this exact instance from the query
+        if instance.pk:
+            found = found.exclude(pk=instance.pk)
```
anastasia, line 24: Why do we need this filter? Shouldn't this only ever be called when there is no slug, therefore instance will never be in the queryset?

anastasia: Nope! I take it back. I see that we want to be able to call this whenever, like in the case of jurisdiction slug setting.

capstone/capdb/utils.py
jcushman: My gut feeling is the :param: thing isn't necessary -- most params are self-explanatory (or can be with good names), we're not generating HTML documentation, and we're unlikely to document :params: consistently. I feel like the time would be better spent documenting how to call the function or adding other comments. But maybe this is just laziness on my part, could be talked into it.

anastasia: Nah, that makes total sense. I like the conversational aspect of this anyway — and, to be honest, simply switched from this kind of thing (though I wasn't great about writing docstrings to begin with) to the other because of PyCharm autofill.

PR #89, opened 2017-10-16
Download functionality

anastasia:


capstone/capapi/resources.py
```diff
@@ -27,19 +31,54 @@ def format_filename(case_id, whitelisted=False):
...
 
 
-def gzip_documents(zipname, filenames):
-    import zipfile
+def write_and_zip(case_list):
+    def get_matching_case_xml(case_id):
+        try:
+            xml = CaseXML.objects.get(case_id=case_id)
+            return xml.orig_xml
+        except ObjectDoesNotExist:
```
jcushman: Random tip: I usually do these as `CaseXML.DoesNotExist`, which is a little clearer to read and avoids the extra import.

capstone/capapi/resources.py
```diff
@@ -27,19 +31,54 @@ def format_filename(case_id, whitelisted=False):
...
+            logger.error("Case id mismatch", case_id)
+
+    case_xml_tuples = [(case.slug, get_matching_case_xml(case.case_id)) for case in case_list]
+
+
+    dirname_partial = '{0}_{1}'.format(case_list[0].slug[:20], case_list[len(case_list)-1].slug[:20]) if len(case_list) > 1 else case_list[0].slug
```
jcushman: This line is a bit confusing -- maybe add a comment to explain what it's for? It might also help to split into a couple lines:

```
# name download file based on single case, or first and last case if multiple
dirname_partial = case_list[0].slug[:20]
if  len(case_list) > 1:
    dirname_partial += '_' + case_list[-1].slug[:20]
```

capstone/capapi/resources.py
```diff
@@ -27,19 +31,54 @@ def format_filename(case_id, whitelisted=False):
...
+
+    dirname = "{0}_{1}".format(
+        dirname_partial,
+        slugify(datetime.now().timestamp()))
+
+    dirpath = settings.API_ZIPFILE_DIR + "/" + dirname
+
     try:
```
jcushman: I think you just want `zipfile.ZIP_DEFLATED` here, unless there's some reason it would sometimes be undefined?

capstone/capapi/resources.py
```diff
@@ -27,19 +31,54 @@ def format_filename(case_id, whitelisted=False):
...
         compression = zipfile.zip_deflated
-    except:
+    except AttributeError:
         compression = zipfile.ZIP_STORED
 
-    with zipfile.ZipFile(zipname, 'w') as zapfile:
-        for f in filenames:
-            fname = f.split('/')[-1]
-            zapfile.write(f, fname, compress_type=compression)
+    os.makedirs(dirpath)
```
jcushman: Rather than create an actual directory, I think we can do this with a SpooledTempFile. Something like this:

```
tmp_file = tempfile.SpooledTemporaryFile(10*2**20)  # tmp file backed by RAM up to 10MB, then stored to disk
with zipfile.ZipFile(tmp_file, 'w', zipfile.ZIP_DEFLATED) as archive:
    for case in case_list:
        archive.writestr('somepath.txt', get_matching_case_xml(case.case_id))

# Reset file pointer
tmp_file.seek(0)

# return open file handle
return tmp_file

# in your view:
wsgiref.util import FileWrapper
from django.http import FileResponse

return FileResponse(FileWrapper(tmp_file))
```

I know we talked a bit about RAM usage for this process. The general idea is that we want to build the zip file in RAM, for speed, if we can; and we want to avoid keeping too many copies in RAM at the same time, so avoid building a list of XML files and then iterating through it.

I *think* this approach will also cause the temp files to be automatically deleted after delivery; if not we may need to add on a cleanup task for that.

capstone/capapi/resources.py
```diff
@@ -27,19 +31,54 @@ def format_filename(case_id, whitelisted=False):
...
+    # zip newly created directory
+    zipped_dir = dirpath + ".zip"
+    zipfile_handler = zipfile.ZipFile(zipped_dir, 'w', compression)
+    zipdir(dirpath, zipfile_handler)
+    zipfile_handler.close()
+
+    return zipped_dir
+
+
+def zipdir(path, zipfile_handler):
```
jcushman: ... and then in the SpooledTemporaryFile approach I think zipdir(), write_to_file(), and move_casezip() become unnecessary.

capstone/capapi/tests/test_api.py
```diff
@@ -41,6 +47,118 @@ def test_case(load_parsed_metadata):
...
 
+    case = setup_case()
+    response = c.get("%s/cases/%s/?format=json" % (settings.API_FULL_URL, case.slug))
+    check_response(response)
+    content = response.json()
+    assert content.get("name_abbreviation") == case.name_abbreviation
+
+
+@pytest.mark.django_db(transaction=True)
+def test_single_case_download():
```
jcushman: Check out https://docs.pytest.org/en/latest/fixture.html and capdb/tests/conftest.py to see how to make these tests a little simpler with fixtures. We should be able to do these as just `def test_single_case_download(auth_user, client, case):` rather than explicitly set up the same stuff in each test.

(You might want those fixtures to just wrap your existing factory methods, so you can also call the factory methods directly when you want to customize the case or whatever.)

capstone/capapi/view_helpers.py
```diff
@@ -1,35 +1,30 @@
...
-        qwarg = {}
-        if ('_id' not in key) and ('decisiondate' not in key):
-            key = key + '__icontains'
-        qwarg[key] = val
-        return Q(**qwarg)
-
-
-def format_query(params, args_dict):
-    blacklisted_keys = ['type', 'page', 'offset', 'max']
+def format_query(params, args_dict=dict()):
```
jcushman: `args_dict=dict()` isn't safe in Python -- it's always the same dict() every time you call the function!

Instead you can do `args_dict=None`, and then in the function:

```
if args_dict is None:
    args_dict = {}
```

capstone/capapi/view_helpers.py
```diff
@@ -1,35 +1,30 @@
...
-            args_dict['decisiondate__gte'] = val
+            args_dict['decision_date__gte'] = val
         elif key == 'max_year':
-            args_dict['decisiondate__lte'] = val
-        elif '_' in key:
+            args_dict['decision_date__lte'] = val
+        elif '_' in key and key != 'docket_number':
             splitkey = key.split('_')
             args_dict[splitkey[0] + '__' + splitkey[1]] = val
         else:
```
jcushman: For this function, let's not accept arbitrary keys and pass them into Django query `filter()` -- it's hard to be confident about whether that's safe. Let's have a whitelist of query parameters that are allowed for each endpoint. (Switching to built-in DRF filters would do this.)

capstone/capapi/view_helpers.py
```diff
@@ -38,11 +33,43 @@ def format_query(params, args_dict):
...
-def format_url_from_case(case):
-    return "%s/%s/%s/%s/%s" % (
-        case.jurisdiction,
-        case.reporter,
-        case.volume,
-        case.firstpage,
-        case.name_abbreviation
-    )
+def make_query(query_pair):
+    (key, val) = query_pair
```
jcushman: What does this do if val is false? Maybe add an explicit `return None` at the bottom and an explanation of why it's OK (if it is).

capstone/capapi/view_helpers.py
```diff
@@ -38,11 +33,43 @@ def format_query(params, args_dict):
...
+def merge_filters(q_list, operation):
+    reducer_operation = operator.or_ if operation == 'OR' else operator.and_
+    return reduce(reducer_operation, q_list)
+
+
+def generate_filters_from_query_params(query_params):
+    # format queries
+    query_dict = format_query(query_params, dict())
+    # create Q object query dict
+    queries = list(map(make_query, query_dict.items()))
```
jcushman: This might be clearer as `queries = [make_query(k,v) for k,v in query_dict.items()]` (and simplify the make_query() function a bit).

capstone/capapi/view_helpers.py
```diff
@@ -1,35 +1,30 @@
+import logging
```
jcushman: I have some bikeshedding comments below of the query filter stuff. Another general suggestion is to switch to DRF's built-in filtering classes instead of rolling our own for this. Perma has an example here:

https://github.com/harvard-lil/perma/blob/develop/perma_web/api/views.py#L28

We could add that as a bug for later if you don't want to mess with it now, though.

capstone/capapi/views.py
```diff
@@ -56,19 +61,118 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
+    permission_classes = (permissions.IsAPIUser,)
     filter_class = filters.CaseFilter
 
+    def download_many(self):
+        all_cases = self.queryset.all()
+        # get max number of cases to download from query
+        # set max number to daily max if unspecified
+        max_num = int(self.request.query_params.get('max', settings.API_CASE_DAILY_ALLOWANCE))
+
+        try:
```
jcushman: - What does it mean if this raises TypeError? Maybe comment.
- Can simplify by moving the .order_by up to `all_cases = self.queryset.order_by('decision_date')`

capstone/capapi/views.py
```diff
@@ -56,19 +61,118 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
+            to_filter = view_helpers.generate_filters_from_query_params(self.request.query_params)
+            cases = all_cases.filter(to_filter).order_by('decision_date')
+        except TypeError:
+            cases = all_cases.order_by('decision_date')
+
+        if cases.count() == 0:
```
jcushman: I was confused about this, but it makes sense. :) Maybe add a comment like: `# user is requesting a zip but there is nothing to zip, so 404 is the right response. See https://stackoverflow.com/a/11760249/307769`

capstone/capapi/views.py
```diff
@@ -56,19 +61,118 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
+            return JsonResponse({
+                'message': 'Request did not return any results.'
+            }, status=404, )
+
+        # get page number for paginated requests
```
jcushman: Rather than roll your own pagination here, I think you can just do `cases = self.paginate_queryset(cases)`

capstone/capapi/views.py
```diff
@@ -56,19 +61,118 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
+        if max_num <= settings.API_DOWNLOAD_LIMIT:
+            cases = cases[:max_num]
+        else:
+            start_at = (page_num-1) * settings.API_DOWNLOAD_LIMIT
+            end_at = page_num * settings.API_DOWNLOAD_LIMIT
+            cases = cases[start_at:end_at]
+
+        # if request was already paginated
+        whitelisted_filters = view_helpers.get_whitelisted_case_filters()
+        blacklisted_case_count = len(set(cases) & set(all_cases.exclude(whitelisted_filters)))
```
jcushman: This isn't very efficient, right? If we call this endpoint without any filters, `set(all_cases.exclude(whitelisted_filters))` will build a set of all 5 million cases.

Here's my suggestion:

- Add a `whitelisted` boolean to the Jurisdiction model
- Use `all_cases = self.queryset.select_related('jurisdiction')` to make sure they have jurisdiction set without hitting the DB again.
- `blacklisted_case_count = len(c for c in case if not case.jurisdiction.whitelisted)`

capstone/capapi/views.py
```diff
@@ -56,19 +61,118 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
+    def download_one(self, **kwargs):
+        """
+        If downloading a single case (using its lookup_field) explicitly
+        """
+        try:
+            case = models.CaseMetadata.objects.get(**kwargs)
+        except ObjectDoesNotExist as e:
+            return JsonResponse({'message': 'Unable to find case with matching slug: %s' % e}, status=404, )
+
+        blacklisted_case_count = 0 if case.jurisdiction.name in OPEN_CASE_JURISDICTIONS else 1
```
jcushman: See above -- let's do these as `case.jurisdiction.whitelisted`

capstone/capapi/views.py
```diff
@@ -56,19 +61,118 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
+                zip_filename = self.get_zip_filename(case_list)
+
+                # update case allowance
+                self.request.user.update_case_allowance(case_count=blacklisted_case_count)
+
+                response = StreamingHttpResponse(FileWrapper(open(zip_filename, 'rb')), content_type='application/zip')
+                response['Content-Length'] = os.path.getsize(zip_filename)
+                response['Content-Disposition'] = 'attachment; filename="%s"' % zip_filename
+                return response
+            except Exception as e:
```
jcushman: I don't think we want a try-except here -- just show a 500 if something goes wrong.

capstone/capapi/constants.py
```diff
@@ -0,0 +1,4 @@
+OPEN_CASE_JURISDICTIONS = [
```
jcushman: See later comments -- I suggest making `whitelisted` a field on the Jurisdiction model instead of setting this as a constant.

jcushman (commented):
Great work! Bunch of comments on this one. :)

capstone/capapi/views.py
anastasia: much better, thanks!

capstone/capapi/views.py
anastasia: TypeError happened if no query_params were passed, setting filters to None. Filtering on None caused an error. I think all of this can be simplified by checking whether filters exist in the first place, and only then filtering the cases.

capstone/capapi/views.py
```diff
@@ -56,19 +57,111 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
+            return JsonResponse({
+                'message': 'Request did not return any results.'
+            }, status=404, )
+
+        cases = self.paginate_queryset(cases)
+
+        blacklisted_case_count = len(list((case for case in cases if not case.jurisdiction.whitelisted)))
+        case_allowance_sufficient = self.check_case_allowance(blacklisted_case_count)
+
+        response = self.create_download_response(list(cases), blacklisted_case_count, permitted=case_allowance_sufficient)
```
jcushman: Nit: paginate_queryset already returns a list, don't need `list(cases)` here

capstone/capapi/views.py
```diff
@@ -56,19 +57,111 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
+        case_allowance_sufficient = self.check_case_allowance(blacklisted_case_count)
 
-### User specific views ###
+        response = self.create_download_response([case], blacklisted_case_count, permitted=case_allowance_sufficient)
+
+        return response
+
+    def create_download_response(self, case_list, blacklisted_case_count, permitted=False):
+        if permitted:
+            try:
```
jcushman: We don't want a try-except here, do we? If there's something specific that throws errors we should handle it, otherwise let it throw a 500 so we find out.

capstone/capapi/views.py
```diff
@@ -56,19 +57,111 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
+                from wsgiref.util import FileWrapper
+                from django.http import FileResponse
+
+                zip_filename = resources.create_zip_filename(case_list)
+
+                streamed_file = resources.create_zip(case_list)
+                response = FileResponse(FileWrapper(streamed_file), content_type='application/zip')
+
+                # update case allowance if response ok
+                if response.status_code == 200:
```
jcushman: response.status_code will always be 200, right? I'm assuming it's just a default property of FileResponse.

capstone/config/settings/settings_base.py
```diff
@@ -172,8 +176,9 @@
 API_FULL_URL = os.path.join(API_BASE_URL_ROUTE, API_VERSION)
 API_CASE_FILE_TYPE = '.xml'
 
-# CAP API EMAIL #
+API_ZIPFILE_DIR = '/tmp'
```
jcushman: No longer used?

capstone/capapi/views.py
```diff
@@ -56,19 +57,111 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
 
+    def download_many(self):
+        cases = self.queryset.all().order_by('decision_date')
+
+        for backend in list(self.filter_backends):
+            cases = backend().filter_queryset(self.request, cases, self)
+
+        # get max number of cases to download from query
+        # set max number to daily max if unspecified
+        cases = cases[:int(self.request.query_params.get('max', settings.API_CASE_DAILY_ALLOWANCE))]
```
jcushman: I think this line might be replaced by pagination?

capstone/capapi/resources.py
```diff
@@ -1,50 +1,45 @@
...
 
+def create_zip_filename(case_list):
+    ts = slugify(datetime.now().timestamp())
+    if len(case_list) == 1:
+        return case_list[0].slug + '-' + ts + '.zip'
 
-def move_casezip(filename):
-    new_dest = "%s/%s" % (settings.CASE_ZIP_DIR, filename)
-    os.rename(filename, new_dest)
+    return '{0}_{1}_{2}.zip'.format(case_list[0].slug[:20], case_list[len(case_list)-1].slug[:20], ts)
```
jcushman: Nit: `case_list[len(case_list)-1]` is simpler as just `case_list[-1]`

capstone/capapi/views.py
```diff
@@ -56,19 +57,111 @@ class CitationViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins
...
 
-### User specific views ###
+        response = self.create_download_response([case], blacklisted_case_count, permitted=case_allowance_sufficient)
+
+        return response
+
+    def create_download_response(self, case_list, blacklisted_case_count, permitted=False):
+        if permitted:
+            try:
+                from wsgiref.util import FileWrapper
```
jcushman: Nit: imports at top?

jcushman (commented):
Nice, just a few more :)

capstone/capapi/resources.py
anastasia: oh my. Brainfart.

capstone/capapi/views.py
anastasia: Actually the thought here was to allow the user to specify how many cases they want, instead of getting all possible for a certain query. Something like, get me the first case, or ten cases.

jcushman: That makes sense, but I think they can do it through pagination. Not sure how you have pagination configured, but it would be something like `http://api.example.org/accounts/?page=4&page_size=100`. So `http://api.example.org/accounts/` will get the first 100, `http://api.example.org/accounts/?page_size=10` will get the first 10, and so on. It's a bit less intuitive, but since you have to figure out pagination to work with any API anyway I think it's usable enough.

(Note that `paginate_queryset` is implemented the same as your line here -- it does a `query = query[start:end]`, which I'm not sure how the Django ORM will interpret the second time. The code now asks for two limit clauses in the sql query, which doesn't make sense. So if we do want to keep both we have to figure out how they should interact.)

anastasia: Aha! Yes. Changed to using DRF's LimitOffsetPagination — I find "limit" to be much more intuitive than "page_size".

PR #91, opened 2017-10-19
add - make it easier for users to browse cases

anastasia:


ChefAndy:
Nice changes. Makes for a friendlier API experience.

PR #93, opened 2017-10-19
Inventory testing

jcushman:
So I started pulling this thread yesterday ("I wonder if I can generalize how we handle storages in capstone ...") and got farther than I expected. :) Stuff in this pull request:

- `settings.STORAGES` becomes a generic set of storage configs. So you can configure `settings.STORAGES['inventory_storage']` to point to some set of files located on S3 or the local disk or whatever, and then do `from capdb.storages import inventory_storage` and it's automatically defined.

- Add a `fab write_inventory_files` command that inspects what's in test_data/from_vendor, and writes an S3-style inventory report to test_data/inventory/data.

- Update ingest_by_manifest.py to use ingest_storage and inventory_storage instead of raw S3 calls.

- Configure STORAGES in settings_base.py so `fab total_sync_with_s3` ingests from test_data/ instead of from S3.

- Update our test fixtures to load data from total_sync_with_s3 (pointed at local test_data) instead of the old ingest script.

- Move redis_client setup to capdb.storages. (I verified that it's safe to do this in multiprocessing -- each process ends up with its own redis connection because of some sort of lazy loading the redis library does.)

- A little refactoring/extra helper functions in ingest_by_manifest that I added while tracing through to figure out why my storages change was breaking stuff.


ChefAndy:
Nice edits and cleanup here. Looks great.

ChefAndy (commented):
I think this is all great. Cleaner, more readable, and more uniform.

PR #95, opened 2017-10-24
Updates volumes instead of deleting w/new version

ChefAndy:
Changed ingest_by_manifest: instead of deleting a volume and recreating
with the new volume, it updates the existing volume.

### DIFF
--- a/capstone/scripts/ingest_by_manifest.py
+++ b/capstone/scripts/ingest_by_manifest.py
@@ -227,11 +227,9 @@ def process_volume(vol_entry_bytestring):
                     page.cases.set(alto_barcode_to_case_map[alto_barcode])
 
             for spare_case in existing_case_ids:
-                print("Whoa! Spare Case! {}".format(spare_case))
                 r.sadd("spare_cases_{}".format(vol_entry['barcode']), spare_case)
 
             for spare_page in existing_page_ids:
-                print("Whoa! Spare Page! {}".format(spare_page))
                 r.sadd("spare_pages_{}".format(vol_entry['barcode']), spare_page)
 
     except IntegrityError as e:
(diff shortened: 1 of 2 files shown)


capstone/scripts/ingest_by_manifest.py
```diff
@@ -184,39 +184,54 @@ def process_volume(vol_entry_bytestring):
...
-            volume = VolumeXML(orig_xml=ingest_storage.contents(volmets_path), barcode=vol_entry['barcode'], s3_key=volmets_path)
-            volume.save()
+            else:
+                volume = VolumeXML(barcode=vol_entry['barcode'])
 
+            volume.orig_xml = ingest_storage.contents(volmets_path)
+            volume.s3_key = volmets_path
+            volume.save()
+            
+            existing_case_ids = set(CaseXML.objects.filter(volume=volume).values_list('case_id', flat=True))
```
jcushman, line 194: Nit: this is more idiomatic in Django if you use `volume.case_xmls`, which means the same thing as `CaseXML.objects.filter(volume=volume)` but is easier to read. In general if you have one object loaded in and want to fetch related objects, it's more natural in Django to use the first object and work from there.

(See https://docs.djangoproject.com/en/dev/topics/db/queries/#related-objects if you want to dig in more on object relationships in Django ...)

capstone/scripts/ingest_by_manifest.py
```diff
@@ -184,39 +184,54 @@ def process_volume(vol_entry_bytestring):
...
+                case.orig_xml = ingest_storage.contents(case_s3_key)
                 case.save()
                 case.update_case_metadata()
 
                 # store case-to-page matches
                 for alto_barcode in set(re.findall(r'file ID="alto_(\d{5}_[01])"', case.orig_xml)):
                     alto_barcode_to_case_map[vol_entry['barcode'] + "_" + alto_barcode].append(case.id)
-
+            
+            existing_page_ids = set(PageXML.objects.filter(volume=volume).values_list('barcode', flat=True))
```
jcushman, line 212: Ditto nit: `volume.page_xmls` instead of `PageXML.objects.filter(volume=volume)`

jcushman (commented):
Looks good! Just a quick style tweak.

PR #97, opened 2017-10-25
Add add_test_case

### LINKED ISSUE #80: Allow devs to load up arbitrary real data
Right now there is a painstaking process involved in loading up case data. 
Instead, in the near future, it might be worth considering adding a script that would do a few things: 
first off, load up the tracking tool data into the dev's local db
secondly, maybe hit up the prod api (using dev's api keys) to download the necessary case xml
thirdly, run ingest_files scripts to make entries for the newly loaded cases

jcushman:
This adds a fab task to add cases to our test data. Example usage:

`fab add_test_case:32044057892259_0001`

Result is:

- write volume-level files, case file, and associated alto and image files to test_data/from_vendor
- edit volmets so it doesn't include missing files (which messes up import script otherwise)
- update tracking_tool.json so fake tracking tool database includes all necessary metadata

Requires `DATABASES['tracking_tool']` and `STORAGES['ingest_storage']` to be temporarily pointed at live sources in settings.py.

Fixes #80


capstone/scripts/helpers.py
```diff
@@ -117,3 +126,13 @@ def parse_xml(xml):
         xml = xml.encode('utf8')
         
     return PyQuery(xml, parser='xml', namespaces=nsmap)
+
+def copy_file(from_path, to_path, from_storage=None, to_storage=None):
+    """
+        Copy contents of from_path to to_path, optionally using storages instead of filesystem open().
+    """
+    from_open = from_storage.open if from_storage else open
```
anastasia, line 134: I like this a lot!

anastasia: Actually, why do we have to open the file in order to copy from/to? Why not just use `shutil.copyfile`?

jcushman: This is mostly for copies involving Django storages (which might be on local disk or S3 or SFTP or anywhere else) -- it's not necessary if both paths are on disk.

(In general we do want to be using Django storages any time we touch files, though, unless we're specifically dealing with temp files -- that lets the same code run in different kinds of deployments.)

PR #99, opened 2017-11-02
add - documentation around all dbs needed

anastasia:


README.md
```diff
@@ -31,6 +31,12 @@ Clone:
 
     $ git clone https://github.com/harvard-lil/capstone.git
     
+You will need MySQL, Redis, and Postgres for Capstone.
+- MySQL installation:
+  `brew install mysql` or [with apt-get](https://www.digitalocean.com/community/tutorials/how-to-install-mysql-on-ubuntu-14-04)
```
anastasia, line 36: didn't see any use to linking to the official mysql download page (since it's super obfuscated and weird) so added these instead.

ChefAndy: My $0.02— I'd rather leave it more generic than adding something so platform specific.

ChefAndy: Oh, didn't see the "or with apt-get"

PR #105, opened 2017-11-17
add - tqdm to see case parsing progress, printing exceptions

anastasia:
I don't know what the proper handling of errors at this level is: we shouldn't ever get any errors, and so currently I'm catching the mysterious "Exception" so as to not break out of that for loop if something goes wrong. 
We should have a more organized error logging process though probably.


jcushman:
I think catch-and-print is fine for this kind of unexpected error for now. One tweak though -- could you print what case is affected, to make for easier debugging?

PR #109, opened 2017-11-28
Capapi - log requests

anastasia:
added middleware for intercepting all capapi requests, logging light details + speed

### DIFF
--- a/capstone/capapi/views.py
+++ b/capstone/capapi/views.py
@@ -33,28 +33,28 @@ class JurisdictionViewSet(BaseViewMixin, mixins.RetrieveModelMixin, mixins.ListM
     renderer_classes = (renderers.BrowsableAPIRenderer, renderers.JSONRenderer)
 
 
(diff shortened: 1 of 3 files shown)


capstone/capapi/views.py
```diff
@@ -28,35 +33,35 @@ class JurisdictionViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mi
     renderer_classes = (renderers.BrowsableAPIRenderer, renderers.JSONRenderer)
 
 
-class VolumeViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins.ListModelMixin,):
+class VolumeViewSet(BaseViewMixin, viewsets.GenericViewSet, mixins.RetrieveModelMixin, mixins.ListModelMixin):
```
ChefAndy: Do you need to have viewsets.GenericViewSet as an argument here when you've got BaseViewMixin? You don't have it in JurisdictionViewSet but you've got it in the rest. (I'm not super familar with your code so I dunno... just stuck out)

anastasia: good catch!

PR #111, opened 2017-12-04
Celery (and docker)

jcushman:
- Add celery integration
- Add test_slow celery task that just wastes a bunch of RAM and CPU
- Add celery_resource_autoscaler integration to scale celery workers to available RAM and CPU
- Add simple docker config to run rabbitmq for testing


bensteinberg (approved):
Looks good to me. Let's discuss what we need for deployment.

PR #117, opened 2017-12-06
Celery

anastasia:
added:
- celery
- case metadata writing celery task
- fab task to call (method to call said) task


anastasia:
ha ha oh no there's already a tasks.py — should have pulled the latest changes

PR #119, opened 2017-12-11
Xml lazy load

jcushman:
- Abstract common functionality from VolumeXML/PageXML/CaseXML into BaseXMLModel.
- Add BaseXML.objects.defer_xml() function, which is equivalent to BaseXML.objects.defer('orig_xml'). (This isn't a huge win, but it's a bit handy I guess.)
- Use memory-efficient queries in a couple of places where we iterate over XML tables.
- Drop the old unused update_case_metadata.

### DIFF
--- a/capstone/capdb/models.py
+++ b/capstone/capdb/models.py
@@ -13,19 +13,41 @@ def choices(*args):
     return zip(args, args)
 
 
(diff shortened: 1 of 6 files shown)


anastasia:
👍

PR #120, opened 2017-12-11
Case allowance

anastasia:
while we don't have an idea for a structure for different types of users, we can use the `total_case_allowance` field on the APIUser model to add custom case allowance. This means that every time 24 hours passes since the "case_allowance_last_updated" field, that user's case_allowance won't revert to the `settings.API_CASE_DAILY_ALLOWANCE` default, but instead refresh to whatever that particular user's `total_case_allowance` is. 
to that end, I changed the `case_allowance` name to be `case_allowance_remaining`


capstone/capapi/admin.py
```diff
@@ -18,17 +20,16 @@ def authenticate_user(modeladmin, request, queryset):
...
-        'case_allowance',
-        'date_joined',
+        'first_name',
         'api_key',
-        'is_researcher',
+        'case_allowance_remaining',
+        'total_case_allowance',
     )
 
+    fields = deepcopy(list_display) + ('is_active', 'date_joined', 'activation_nonce', 'is_researcher')
```
jcushman: Seems like this works the same without deepcopy?

jcushman (commented):
Cool! Just one nitpick about deepcopy.

capstone/capapi/admin.py
anastasia: yes why it absolutely does

PR #124, opened 2017-12-14
Nightmare pull request

jcushman:
This should have been about four separate things, but I failed to break it up well while working on it. Basically I wanted to run create_or_update_metadata() efficiently, and pulled out way more of the string than I should have in one go.

High-level changes:

- Add a duplicative case to our test data (i.e. one that only has raw OCR from Innodata, with nothing useful in case xml) and fix a couple of edge cases that popped up when including that.
- Add foreign keys from CaseXML.metadata -> CaseMetadata and VolumeXML.metadata -> VolumeMetadata. Barcodes are now stored only on the metadata models, not duplicated on the xml models.
- Change how we generate slugs, so models optimistically set their slugs and attempt to save, then retry with new slugs on failure. This is handled by a model mixin called AutoSlugMixin.
- Add another model mixin called CachedLookupMixin with a method get_from_cache for efficient access to small, rarely-changing models like Court and Jurisdiction.
- Rewrite CaseXML.create_or_update_metadata() using foreign keys and new model mixins for efficiency.
- Tweak a couple of other places that depended on the way foreign keys or slug generation previously worked.
- Refactor tests so we have one set of fixtures usable from both capdb and capapi tests.
- Update tests for new slug handling.
- Add test for create_or_update_metadata fab task.


capstone/capdb/migrations/0013_auto_20171211_2010.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.1 on 2017-12-11 20:10
+from __future__ import unicode_literals
```
anastasia, line 3: I think there might be a merge conflict because there already is a 0013_migration

capstone/capdb/migrations/0016_merge_20171214_1624.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('capdb', '0015_auto_20171212_1749'),
+        ('capdb', '0013_auto_20171213_1548'),
+    ]
+
+    operations = [
```
anastasia, line 15: what's the need for this migration?

anastasia: this is *not* in fact the 0013 migration from your pull request, so that makes total sense now

PR #125, opened 2017-12-14
Don't drop case_id field yet

jcushman:
Oops -- for deployment, let's not do RemoveField yet, so we can confirm the new field was populated correctly.


bensteinberg (approved):
LGTM!

PR #128, opened 2017-12-15
Change first_page and last_page to charfields

jcushman:
Some <casebody first_page= and last_page=> are not integers. Let's just store the straight strings for now -- we may want to add an ordinal integer later. This will fix some CaseMetadata imports that currently aren't working.


jcushman:
Update -- also increasing length of CaseMetadata.docket_number, which is blocking some other cases from importing.

PR #129, opened 2017-12-18
Updated CaseXML.save to check for ALTO updates

ChefAndy:
modified:   capdb/models.py - added new save method, new init method to
                              make a copy of the orig_xml when the object
                              is instantiated, and modified the
                              create_or_update_metadata method with a
                              kind of inelegant fix for when a case is
                              loaded with orig_xml deferred, and then
                              loaded and saved.
modified:   capdb/tests/test_models.py - added a bunch of tests for the new
                                         methods


ChefAndy:
I have no idea why the functionality tested by test_casebody_modify_word isn't working. It's working on my system locally. Python versioning shenanigans? I'll check tomorrow.

capstone/capdb/models.py
```diff
@@ -421,13 +458,153 @@ class CaseXML(BaseXMLModel):
...
+                               on_delete=models.DO_NOTHING)
     s3_key = models.CharField(max_length=1024, blank=True, help_text="s3 path")
 
     tracker = FieldTracker()
 
+    def save(self, force_insert=False, force_update=False, *args, **kwargs):
+
+        # This compares new XML to the old and updates ALTO files as necessary.
+        # We don't yet support adding or removing words from the case body.
+        if self.tracker.has_changed('orig_xml') and self.pk is not None and self.tracker.changed()['orig_xml']:
```
jcushman: Maybe more readable: `if self.tracker.has_changed('orig_xml') and self.tracker.previous('orig_xml'):`?

capstone/capdb/models.py
```diff
@@ -421,13 +458,153 @@ class CaseXML(BaseXMLModel):
...
+            def __in_casebody(element):
+                return element.tag.startswith("{" + nsmap['casebody']) and not element.tag.endswith('casebody')
+
+            # parse objects and get the xml tree objects
+            parsed_orig_case = parse_xml(self.tracker.changed()['orig_xml'])
```
jcushman: `self.tracker.changed()['orig_xml']` -> `self.tracker.previous('orig_xml')`

capstone/capdb/models.py
```diff
@@ -421,13 +458,153 @@ class CaseXML(BaseXMLModel):
...
+            parsed_updated_case = parse_xml(self.orig_xml)
+            updated_tree = parsed_updated_case.root
+            original_tree = parsed_orig_case.root
+
+            # If the case is duplicative, just save and move on
+            if updated_tree.find('//duplicative:casebody', nsmap) is not None:
+                super(CaseXML, self).save(force_insert, force_update, *args, **kwargs)
+                return
+
+            if self.case_id == '':
```
jcushman: I think we're headed for just keeping CaseMetadata.case_id, which should always be set. Maybe just drop this?

capstone/capdb/models.py
```diff
@@ -540,13 +717,27 @@ def __str__(self):
...
+    tracker = FieldTracker()
+
+    def save(self, force_insert=False, force_update=False, *args, **kwargs):
+        super(PageXML, self).save(force_insert, force_update, *args, **kwargs)
+        if self.tracker.has_changed('orig_xml') and self.pk is not None and self.tracker.changed()['orig_xml']:
+            split_barcode = self.barcode.split('_')
+            short_alto_id = "alto_{}_{}".format(split_barcode[1], split_barcode[2])
+            self.volume.update_related_md5(short_alto_id, self.md5())
+            for case in self.cases.all():
+                split_barcode = self.barcode.split('_')
```
jcushman: This is unfinished, right? Should update case instead of volume.

capstone/capdb/models.py
```diff
@@ -540,13 +717,27 @@ def __str__(self):
...
     barcode = models.CharField(max_length=255, unique=True, db_index=True)
     volume = models.ForeignKey(VolumeXML, related_name='page_xmls',
-                                     on_delete=models.DO_NOTHING)
+                               on_delete=models.DO_NOTHING)
     cases = models.ManyToManyField(CaseXML, related_name='pages')
     s3_key = models.CharField(max_length=1024, blank=True, help_text="s3 path")
 
+    tracker = FieldTracker()
+
+    def save(self, force_insert=False, force_update=False, *args, **kwargs):
```
jcushman: This gets a bit recursive -- we'll have a Case save triggering a Page save triggering a Case save. Can we do this so each thing is only saved once? 

Maybe by adding options for it -- `update_case_md5 = kwargs.pop('update_case_md5'); update_volume_md5 = kwargs.pop('update_volume_md5');`. Then in Case.save() you can pass in False, and take care of it there. This would probably also mean adding a `save=True` option to `update_related_md5()` so you can update in Case.save without immediately saving.

capstone/capdb/models.py
```diff
@@ -421,13 +458,153 @@ class CaseXML(BaseXMLModel):
...
+
+            # If the case is duplicative, just save and move on
+            if updated_tree.find('//duplicative:casebody', nsmap) is not None:
+                super(CaseXML, self).save(force_insert, force_update, *args, **kwargs)
+                return
+
+            if self.case_id == '':
+                self.case_id = parsed_orig_case('case|case').attr('caseid')
+
+            # compare a flat list of xpaths to see if there are structural changes
```
jcushman: How much of this (lines 491 to 521) is still necessary? I don't think we need to edit any trees, for example.

capstone/capdb/models.py
```diff
@@ -421,13 +458,153 @@ class CaseXML(BaseXMLModel):
...
+
+            # get the alto files associated with the case in the DB
+            alto_files = {}
+            for alto in self.pages.all():
+                alto_fileid = "_".join((["alto"] + alto.barcode.split('_')[1:3]))
+                alto_files[alto_fileid] = alto
+
+            # check to see if any elements in the casebody have been updated so we can update the ALTO
+            # since the tree structures match, we can just iterate over the tree elements and compare values
+            for original_element in original_tree.iter():
```
jcushman: Faster to do this as `original_tree('casebody|casebody')` instead of iterating all and skipping?

PR #140, opened 2017-12-20
Add fab task to fix xml columns and add md5s

jcushman:


capstone/capdb/tasks.py
```diff
@@ -43,3 +44,19 @@ def test_slow(i, ram=10, cpu=30):
...
+@shared_task
+@transaction.atomic
+def fix_md5_column(volume_id):
+    """
+        Our database has xml fields in the casexml and pagexml tables that are missing the <?xml> declaration, and that also don't have the md5 column filled.
+
+        Here we update all the xml fields to add the <?xml> declaration and md5 hash.
+    """
+    with connection.cursor() as cursor:
+        new_xml_sql = "E'<?xml version=''1.0'' encoding=''utf-8''?>\n' || orig_xml"
```
bensteinberg, line 58: Not sure I see how this prepends the line if necessary -- maybe I'm not getting the syntax here.

jcushman: Once all the string munging is done, the task ends up running this SQL, which is maybe easier to read than the python:

```
UPDATE capdb_casexml SET orig_xml=xmlparse(CONTENT E'<?xml version=''1.0'' encoding=''utf-8''?>\n' || orig_xml), md5=md5(E'<?xml version=''1.0'' encoding=''utf-8''?>\n' || orig_xml) where volume_id = 3 and md5 is null;
UPDATE capdb_pagexml SET orig_xml=xmlparse(CONTENT E'<?xml version=''1.0'' encoding=''utf-8''?>\n' || orig_xml), md5=md5(E'<?xml version=''1.0'' encoding=''utf-8''?>\n' || orig_xml) where volume_id = 3 and md5 is null;
```

So it prepends the line only if md5 is null, which corresponds in our database to exactly the entries that need it. And it updates md5 so running again won't do anything.

(We shouldn't run this willy-nilly in the future -- once we get it run for all rows, we can delete it, and md5=null won't have any special significance.)

bensteinberg: Thanks. The double pipe concatenation was tripping me up.

PR #143, opened 2018-01-02
ingest_by_manifest refactor

jcushman:
Some refactoring of ingest_by_manifest for performance.

- Ingest becomes a celery task. Instead of running a process pool, parallel work goes into subtasks handled by individual celery workers.
- More efficient redis use. We now have a "volumes" queue that just lists the name of each volume folder, a "volume:<volume_folder>" queue per volume folder, and an "errors" queue. Entries in a given inventory file are packed together so `read_inventory_file` only needs two redis calls per volume folder.
- Only validate the volmets against s3 file listing if we're fetching and updating volumexml.
- Optimize `ingest_volume` to run the minimum number of sql queries. Prefetch and cache all necessary data, only save if something changed.

I moved some things around in the process, so the code now pretty much reads from top to bottom, which made it easier for me to see any performance issues as I went along.

### DIFF
--- a/capstone/capdb/models.py
+++ b/capstone/capdb/models.py
@@ -5,8 +5,9 @@ from django.utils.text import slugify
 from django.utils.encoding import force_bytes
 from model_utils import FieldTracker
 
+from scripts.helpers import special_jurisdiction_cases, jurisdiction_translation
 from scripts.process_metadata import get_case_metadata
-from scripts.helpers import *
+
 
 ### helpers ###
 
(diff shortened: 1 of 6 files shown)


ChefAndy:
Looks like redis isn't starting in the test env or something like that. All of the tests pass locally for me.

bensteinberg:
Do we need

     services:
      - redis-server

in `.travis.yml`?

jcushman:
Sorry, fixed! We patch redis in a pytest fixture to (a) launch and (b) reset between each test, but I forgot to patch the second database connection that I added in this pull for use by the ingest script.

ChefAndy:
Nice. Ingest_by_manifest looks much cleaner.

PR #150, opened 2018-01-10
Allow sync subtasks in celery

jcushman:
celery can you just be cool for one minute


bensteinberg (commented):
Think we need a more ominous alias for `disable_sync_subtasks`? LGTM.

Issue #127, opened 2017-12-15
Drop unused parsed_metadata.json file?


jcushman:
I think parsed_metadata.json and the fixture to import it are unused -- should we drop them?

Issue #88, opened 2017-10-03
Document fab tasks


jcushman:
Our fab tasks need better names and docstrings, so `fab -l` prints something useful. Right now we have:

```
$ fab -l
Available commands:
    sync_metadata
    update_case_metadata
```

... which is confusing to my brain.

Issue #87, opened 2017-10-03
Give related_name model attributes meaningful values


jcushman:
Right now they're things like `related_name='%(class)s_jurisdiction'`, when they should be things like `related_name='case_metadatas'`.

anastasia:
Add tests to make sure that related_names are set correctly

PR #154, opened 2018-01-11
closing out some old issues

anastasia:
finishes #127 #88  
touches #87 
not seeing other fabtasks that need documenting aside from sync_metadata
added some tests for related_name attr, can keep issue open if you think PageXML and TrackingToolLog models need to be tested as well (this requires writing test models so will take a little bit more time)
dropped parsed_metadata file


jcushman:
Nice, thanks!

PR #156, opened 2018-01-12
Capapi fetching related

anastasia:


capstone/config/settings/settings_dev.py
```diff
@@ -26,3 +26,17 @@
     # this will only be used if CELERY_TASK_ALWAYS_EAGER = False
     CELERY_BROKER_URL = 'redis://redis'
     CELERY_RESULT_BACKEND = 'redis://redis'
+
+
+# turn sql debugging on
+DEBUG_SQL = False
+
+if DEBUG_SQL:
```
ChefAndy: It doesn't seem like there's any scenario in which this code would run— Is it just a placeholder or something?

anastasia: It's been helping me debug (that's why I put it in settings_dev) and it's just a bummer to have to write each time I want to log sql. It's just a switch for on and off, if you need sql logs for any reason, set DEBUG_SQL to True

anastasia: crap, it should probably say "turn sql logging on" instead of "debugging". My bad.

PR #160, opened 2018-02-01
CaseXML->HTML converter and tests

ChefAndy:
new file:   scripts/generate_case_html.py: CaseXML->HTML converter 
modified:   scripts/tests/test_helpers.py: and tests


jcushman:
Aww heck yes.

PR #173, opened 2018-02-21
People in cases fix

anastasia:
take two
this pull request includes moving around some tests to exist in test directories that make more sense


jcushman:
Nice!

PR #175, opened 2018-02-22
removing ingest_files.py and related fab task

anastasia:


jcushman:
Thanks! Sorry about leaving that hanging around.

PR #177, opened 2018-02-22
Create redis_client lazily

jcushman:
I think this *might* solve the celery crash that stops us from creating > 128 celery workers, but I can't confirm locally because docker doesn't have the resources to try it.


bensteinberg (approved):
Let's try it

bensteinberg:
I was unable to run celery with `concurrency=256` with this change in place.

PR #182, opened 2018-02-23
Report results from validation task

jcushman:
This is just for posterity -- I've already run it on the results of our last pass.


bensteinberg (approved):
cool except for flake8

PR #183, opened 2018-02-26
Upgrade to Django 2

jcushman:
- Upgrade Django
- Upgrade DRF and its filtering packages for 2.0 compat
- Tweak a couple of ORM-related things where Django 2 is a little more strict
- Switch "name" to "field_name" for filters as requested by the newest version of django-filters
- Rename a couple of tests


jcushman:
Oops, hold on this, it's messing with the API ...

jcushman:
OK, I think the API issue was unrelated. This should be ready for review.

PR #185, opened 2018-03-01
Citation rearchitecture and serving up some casebody formats

anastasia:
- moved from the inconsistent `/cases/slugified-citation` to `/cases/id`
- allowing users to use the shorthand `/cases/real-citation` and redirecting users to `/cases/?cite=real-citation` which returns a list of cases that match the citation
- allowing users to use normalized cite in the same way: `/cases/normalized-cite` redirects to `/cases/?cite=real-citation`
- restructured the way API permissions are done, see note in code
- removed download functionality
- allowing user to request full_case in browser if case is whitelisted with no api token
- allowing user to get full_case if they have the proper credentials (in near future would be nice to allow passing apikey as param)
- allowing casebody's body_format parameter to be set to xml and html (defaults to text) if `format` param is set to `json` or `api` or is missing (defaults to `api`)

This pull request includes migrations! Should discuss how best to handle them.

This pull request does not cover!
- documentation - if we agree on these approaches I will absolutely rewrite the docs and add a lot more examples. Thought it best not to rewrite them before I know for sure this is correct.
- A big sweep of format requests:
  Not returning a full XML case yet (cases/id/?format=xml), just metadata + xml body. Need to spend a bit more time on this and didn't want to hold this up even more
  Not returning a full HTML case yet (cases/id/?format=html), just metadata + html body. This should be a quick one because @ChefAndy did all the work.


capstone/capapi/permissions.py
```diff
@@ -18,9 +19,41 @@ def has_permission(self, request, view):
...
+        # serve up whether the user is signed in or not
+        if not blacklisted_count:
+            return True
+
+        # from here forth one cannot be anonymous
+        if request.user.is_anonymous:
+            return False
+
+        is_allowed = request.user.case_allowance_remaining >= blacklisted_count
+
```
anastasia: This is logic that used to exist in the view's CaseViewSet download function. It has been moved here because I think permissions logic appropriately should live in the permissions module. I am still conflicted about this, though, because this means that we're doubling the lookups, as DRF's permissions only (I think) allows the returning of a Boolean. This means that if a user *does* have permissions, this look up will happen again in the CaseViewSet. I'm hoping django's and psql's caching might be helpful to speed this up, but I don't know this for sure. What do you think?

capstone/capapi/permissions.py
```diff
@@ -18,9 +19,41 @@ def has_permission(self, request, view):
             # boot if user tries anything but `GET`, `OPTIONS`, or `HEAD`
             return False
 
-        # give away metadata to everyone!
-        if not request.query_params.get('type') == 'download':
+        # check whether user wants to get full case text
+        context = view.get_renderer_context()
+        full_case = context.get('request').query_params.get('full_case', 'false').lower()
```
jcushman: Is `context.get('request')` different from `request`?

capstone/capapi/permissions.py
```diff
@@ -18,9 +19,41 @@ def has_permission(self, request, view):
             # boot if user tries anything but `GET`, `OPTIONS`, or `HEAD`
             return False
 
-        # give away metadata to everyone!
-        if not request.query_params.get('type') == 'download':
+        # check whether user wants to get full case text
```
jcushman: All this stuff should be happening in a special `class CasePermission` rather than `class IsAPIUser`, right? `IsAPIUser` is also used by `resend_verification`, which makes sense, but this all makes more sense if `IsAPIUser` actually is just a test of whether someone is authenticated as an API user.

capstone/capapi/views.py
```diff
@@ -75,64 +76,38 @@ class CaseViewSet(BaseViewMixin, mixins.RetrieveModelMixin, mixins.ListModelMixi
...
-
-        return response
+            return self.serializer_class
 
     def list(self, *args, **kwargs):
-        if not self.request.query_params.get('type') == 'download':
-            return super(CaseViewSet, self).list(*args, **kwargs)
-        else:
-            return self.download(**kwargs)
+        self.serializer_class = self.get_serializer_class(self, *args, **kwargs)
```
jcushman: Don't think this line is necessary -- `self.get_serializer_class` is called automatically by `self.get_serializer`, which is called by `super`.

capstone/capapi/views.py
```diff
@@ -75,64 +76,38 @@ class CaseViewSet(BaseViewMixin, mixins.RetrieveModelMixin, mixins.ListModelMixi
...
+        cite = self.request.query_params.get('cite', None)
+        if cite:
+            self.queryset = self.queryset.filter(citation__normalized_cite=cite)
```
jcushman: Shouldn't this cite section be handled by `CaseFilter`, like the other filters?

(I think it's actually risky to modify `self.queryset` directly -- couldn't that potentially persist between requests?)

capstone/capapi/views.py
```diff
@@ -75,64 +76,38 @@ class CaseViewSet(BaseViewMixin, mixins.RetrieveModelMixin, mixins.ListModelMixi
...
+        return super(CaseViewSet, self).list(*args, **kwargs)
 
     def retrieve(self, *args, **kwargs):
-        if not self.request.query_params.get('type') == 'download':
-            return super(CaseViewSet, self).retrieve(*args, **kwargs)
+        slugified = slugify(kwargs[self.lookup_field])
+        # for user's convenience, if user gets /cases/case-citation or /cases/Case Citation
+        # we redirect to /cases/?cite=case-station
+        if '-' in slugified:
+            query_params = ''.join(list(map(lambda k: '&%s=%s' % (k, self.request.query_params[k]), self.request.query_params)))
```
jcushman: This won't work if values have `&` or `=` in them -- you want to encode the parameters. It's also a good idea to use the reverse() function for the URL. Something like this should work (borrowed from Perma's current_query_string templatetag):

```
import urllib
from rest_framework.reverse import reverse

query_string = urllib.urlencode(dict(self.request.query_params, cite=slugified), doseq=True)
new_url = reverse('api:cases-list', request=self.request) + "?" + query_string
```

capstone/capdb/migrations/0026_auto_20180227_1659.py
```diff
@@ -0,0 +1,63 @@
...
+# Generated by Django 1.11.1 on 2018-02-27 16:59
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+from django.utils.text import slugify
+from scripts.helpers import chunked_iterator
+
+
+def create_normalized_cite(apps, schema_editor):
```
jcushman: I don't think it's worth running these python updates, since we have to reindex all the cases anyway to rebuild the relationships. I think just wipe the data from the citation table before we deploy this.

capstone/capdb/migrations/0027_auto_20180228_2228.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.0.2 on 2018-02-28 22:28
```
jcushman: Maybe just delete both migrations and recreate as a single migration?

capstone/capdb/models.py
```diff
@@ -485,10 +484,10 @@ class CaseMetadata(AutoSlugMixin, models.Model):
     duplicative = models.BooleanField(default=False)
 
     def __str__(self):
-        return self.slug
+        return str(self.pk)
```
jcushman: Maybe return self.case_id? At least it sort of means something.

capstone/capdb/models.py
```diff
@@ -783,18 +778,16 @@ def create_or_update_metadata(self, update_existing=True, save_self=True):
...
 
-class Citation(AutoSlugMixin, models.Model):
+class Citation(models.Model):
     type = models.CharField(max_length=100,
                             choices=(("official", "official"), ("parallel", "parallel")))
     cite = models.CharField(max_length=10000, db_index=True)
     duplicative = models.BooleanField(default=False)
-    slug = models.SlugField(max_length=255, unique=True)
+    normalized_cite = models.SlugField(max_length=255, null=True, db_index=True)
+    case = models.ForeignKey('CaseMetadata', related_name='citation', null=True, on_delete=models.SET_NULL)
```
jcushman: Related name wants to be 'citations', since this is a many-to-one.

capstone/capdb/models.py
```diff
@@ -747,32 +743,31 @@ def create_or_update_metadata(self, update_existing=True, save_self=True):
...
                     court.save()
 
+        case_metadata.save()
+
         ### Handle citations
 
-        citations = list()
+        #  delete existing citations
+        #  create new citations
+        Citation.objects.filter(case_id=case_metadata.id).delete()
```
jcushman: I have a suggestion for how to handle this bit to avoid recreating cites that already exist. I'll send a PR to your PR ...

capstone/capdb/models.py
```diff
@@ -783,18 +778,16 @@ def create_or_update_metadata(self, update_existing=True, save_self=True):
...
 
 
-class Citation(AutoSlugMixin, models.Model):
+class Citation(models.Model):
     type = models.CharField(max_length=100,
                             choices=(("official", "official"), ("parallel", "parallel")))
     cite = models.CharField(max_length=10000, db_index=True)
     duplicative = models.BooleanField(default=False)
-    slug = models.SlugField(max_length=255, unique=True)
+    normalized_cite = models.SlugField(max_length=255, null=True, db_index=True)
```
jcushman: We have cites longer than 255. I think they're all data errors, but maybe increase the size of this field for now.

jcushman (commented):
Nice chunk of work! Some suggestions in comments ...

capstone/capdb/models.py
jcushman: Oh also I think it's a good idea to automatically set normalized_cite from cite on save, if `self.tracker.has_changed('cite')`

capstone/capapi/permissions.py
jcushman: I think maybe we were going to handle this in a different way that avoids this problem. Rather than require login if the request includes blacklisted cases, and refusing to return if they ask for too many, we were going to filter the results to return only allowed items. So like logged-out users get

```
{
    "casebody": null,
    "casebody_missing_reason": "Please log in to retrieve casebody field for Massachusetts."
}
```

And logged in users who have exceeded their daily limit would get a different casebody_missing_reason.

Some issues to kick around with this approach too ...

capstone/capapi/permissions.py
```diff
@@ -18,9 +19,41 @@ def has_permission(self, request, view):
...
+
+        # from here forth one cannot be anonymous
+        if request.user.is_anonymous:
+            return False
+
+        is_allowed = request.user.case_allowance_remaining >= blacklisted_count
+
+        if is_allowed:
+            # if allowed, discount the number of blacklisted cases served to the user
+            request.user.update_case_allowance(case_count=blacklisted_count)
```
jcushman: Oh, thinking about this over the weekend, I think it doesn't work to update case allowance here. For example it doesn't take pagination into account, so we end up using up your entire case allowance even if you only request the first 10 cases. In order to take everything like that into account, you have to manually recreate the whole view logic here -- better to just hook in after it runs.

capstone/capapi/permissions.py
anastasia: oh no it isn't! I lost sight of the arguments passed here, sorry. Will fix.

capstone/capapi/permissions.py
anastasia: The reason I changed this to a more substantive user permissions check was because a user is an API user whether or not they have the API token. I want to provide as much access to the API as possible, and that means checking whether a case is whitelisted and if it is, allowing a user through even when they are anonymous. Maybe there should be a new name for the authenticated user permissions check, like `IsAuthenticatedUser`?

jcushman: That makes sense! The part I'm getting at here is that this permissions class should *only* be used for the case endpoint -- that's the only endpoint where all the logic here applies, right?

You're also using this class for the resend_verification endpoint, which doesn't work. So it makes sense to change a couple of things:

- Name this something so it should clearly only be used from the case endpoint -- `IsAPIUser` sounds too generic for what it is.
- Use a different class for `resend_verification`.

capstone/capapi/permissions.py
```diff
@@ -11,16 +11,50 @@ def has_permission(self, request, view):
...
+        return casebody
+
+    if not case.jurisdiction.whitelisted:
+        if request.user.is_anonymous:
+            casebody["status"] = casebody_permissions[2]
+            return casebody
         else:
-            # make sure user has auth token before proceeding if they want to download
-            return request.auth
+            request.user.update_case_allowance()
```
jcushman: This issues two sql updates per blacklisted case (here and in the else). Can we get it down to one per request?

capstone/capapi/serializers.py
```diff
@@ -6,34 +6,50 @@
...
     court = serializers.ReadOnlyField(source='court.name')
     court_url = serializers.HyperlinkedRelatedField(source='court', view_name='court-detail', read_only=True, lookup_field='slug')
     reporter = serializers.ReadOnlyField(source='reporter.full_name')
     reporter_url = serializers.HyperlinkedRelatedField(source='reporter', view_name='reporter-detail', read_only=True)
-    citations = CitationSerializer(many=True)
+    citations = serializers.SerializerMethodField()
     volume = serializers.ReadOnlyField(source='volume.volume_number')
     volume_url = serializers.HyperlinkedRelatedField(source='volume', view_name='volumemetadata-detail', read_only=True)
 
+    @staticmethod
```
jcushman: Is this different from doing `citations = CitationSerializer()`, as in http://www.django-rest-framework.org/api-guide/relations/#nested-relationships ?

capstone/capapi/permissions.py
```diff
@@ -11,16 +11,50 @@ def has_permission(self, request, view):
...
+
+
+class IsSafeMethodsUser(permissions.BasePermission):
+    def has_permission(self, request, view):
+        # we're a read-only operation here
+        # boot if user tries anything but `GET`, `OPTIONS`, or `HEAD`
+        return request.method in permissions.SAFE_METHODS
+
+
+casebody_permissions = [
```
jcushman: What if we expose `jurisdiction.whitelisted` as its own field? (For example, make `jurisdiction` a nested serializer on the case serializer, so we could also drop the `jurisdiction_url` field.)

Then we could simplify this to four options: ["ok", "error_auth_required", "error_limit_exceeded", "error_unknown"]. And you can check for success with just `status == "ok"`.

capstone/scripts/helpers.py
```diff
@@ -233,6 +233,10 @@ def extract_casebody(case_xml):
         if label and footnote[0].text.startswith(label):
             footnote[0].text = footnote[0].text[len(label):]
 
-    return case('casebody|casebody').html()
+    return case('casebody|casebody')
 
- 
\ No newline at end of file
+
+def extract_casetext(case_xml):
```
jcushman: Unused function?

capstone/test_data/test_fixtures/factories.py
```diff
@@ -25,10 +30,16 @@ def setup_case(**kwargs):
...
 
-### factories ###
+def setup_authenticated_user(**kwargs):
+    user = APIUserFactory.create(**kwargs)
+    token = APITokenFactory.build(user=user)
+    token.save()
+    return user
+
+
+def setup_user(**kwargs):
```
jcushman: Unused fixture?

jcushman (commented):
Looking good! A few more questions ...

capstone/capapi/serializers.py
anastasia: woops, no! I kept sticking to fields instead of going with nested relationships because of the need to obscure what I'd guess is a bunch of unnecessary data in the case list view. But here, we're exposing everything we have to offer for citations anyway, so might as well go with the citation serializer.
I will keep the rest of the fields (minus jurisdiction) as ReadOnlyFields/HyperlinkedFields though — otherwise it gets overly cluttered with information

PR #187, opened 2018-03-12
Add ingest_volume_from_s3 task to directly import without manifest

jcushman:
For our remaining 400 volumes, this lets us directly attempt to ingest each volume instead of waiting for the weekly manifest. The sync_with_s3 task (as opposed to total_sync_with_s3) is updated to use the direct version.


bensteinberg (commented):
LGTM

PR #188, opened 2018-03-13
Model-based versioning

jcushman:
Here's a first pass at low-level versioning using the Django ORM and postgres database triggers.

Here's the explanation I added to the README (consider if this is enough explanation):

> For database versioning we use the Postgres temporal tables approach inspired by SQL:2011's temporal databases.
> 
> See [this blog post](http://clarkdave.net/2015/02/historical-records-with-postgresql-and-temporal-tables-and-sql-2011/)
> for an explanation of temporal tables and how to use them in Postgres. 
> 
> We use [django-simple-history](https://django-simple-history.readthedocs.io/en/latest/) to manage creation, migration,
> and querying of the historical tables.
> 
> Data is kept in sync through the [temporal_tables](https://github.com/arkhipov/temporal_tables) Postgres extension
> and the triggers created in our scripts/set_up_postgres.py file.
> 
> Installing the temporal_tables extension is recommended for performance. If not installed, a pure postgres version
> will be installed by set_up_postgres.py; this is handy for development.

Basically any django model with a `history = TemporalHistoricalRecords()` attribute gets versioning enabled automatically. With VolumeXML, for example, this adds a "sys_period" field that tracks the validity time of the latest record, and creates a VolumeXMLHistory model automatically, which has the identical fields (and future migrations) to VolumeXML, except that `id` is no longer a primary key and `history_id` is. Whenever an update is saved to the VolumeXML table, a postgres trigger automatically saves a copy of the old version to the VolumeXMLHistory table. You can then use `VolumeXML.history` to query the historical records. Historical versions are also exposed in the admin when you go to edit a VolumeXML record.

**Deployment notes** 

For production, we should install the [temporal_tables C extension](https://github.com/arkhipov/temporal_tables), which will be a lot faster than the pure postgres version. This is a little more convenient to do before deployment, so you don't have to go in and delete the pure postgres function.

**Todo**

This doesn't currently have any notion of logging -- you can't optionally log the Django user responsible for the change or a text-based note about the change, and can't retrieve a set of changes other than by querying all updates with the same starting validity time. I'm thinking that's OK to save for a second pass, though.

There's also no code currently to expose prior versions in a useful, human-readable form, e.g. with XML diffs from version to version. Again I think it's OK to save that for later.


bensteinberg (commented):
LGTM, pending changes to keep flake8 happy:
`./capdb/migrations/0026_auto_20180309_1936.py:2:1: F401 'datetime.datetime' imported but unused`
`./capdb/tests/test_versioning.py:5:1: F401 'django.utils.timezone' imported but unused`

jcushman:
Fixed flake8, and rebased to fix merge conflicts.

ChefAndy (approved):
Looks great!

PR #191, opened 2018-03-16
Alto style import

ChefAndy:
I'd really like to test this on a larger number of volumes to see how many get rejected. Maybe we should include this test in the volume validation script.


ChefAndy:
"minor efficiency improvement to the save function in capdb/models.py" should be "minor efficiency improvement to the CaseXML process_updated_xml function in capdb/models.py"

jcushman:
Sweet! Let's land this.

PR #192, opened 2018-03-16
First draft of BagIt bag for case XML

bensteinberg:
see http://gwdev-justinlittman.wrlc.org/bagit.html


bensteinberg:
Is it worth validating the result with e.g. https://github.com/LibraryOfCongress/bagit-python ?

jcushman:
Looks good! I do like the idea of adding test coverage to export a bag and validate it with the LC bagit library.

PR #200, opened 2018-03-21
Add fixture to test query counts

jcushman:
Also drops a redundant test file that must have been copied somehow -- `capapi/tests/test_models.py`


anastasia:
Awesome!!

PR #201, opened 2018-03-21
Tweak case ordering and indexes

jcushman:


jcushman:
(Should drop existing indexes on decision_date on prod before applying.)

bensteinberg:
I was wondering about that -- without a migration, would the model change attempt to create the indexes? Would it fail if they were already present?

PR #204, opened 2018-03-23
First cut at showing slow queries

bensteinberg:
I'm intending to use this like so:

    fab show_slow_queries | curl -X POST -H 'Content-type: application/json' -d @- https://hooks.slack.com/...

The query just gets the top five by time at the moment, but I suppose we could pass that as a parameter. We could run `pg_stat_statements_reset()` at the end of the task, or rely on the default value of `pg_stat_statements.max` (5000) to roll over from day to day, or change the value.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ cache:
 before_install:
   - cd capstone
   - cp config/settings/settings_travis.py config/settings/settings.py
-  - sudo echo shared_preload_libraries = 'pg_stat_statements' >> /etc/postgresql/9.6/main/postgresql.conf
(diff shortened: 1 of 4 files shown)


bensteinberg:
The query comes from https://github.com/NikolayS/postgres_dba which is worth a look.

bensteinberg:
I don't understand why this error only emerged just now, but I was able to get past `ImportError: No module named 'google_compute_engine'` by adding `sudo rm -f /etc/boto.cfg` to `before_install` in `/travis.yml`, per https://github.com/travis-ci/travis-ci/issues/7940.

PR #208, opened 2018-03-27
detect differences in alto/casemets content

ChefAndy:
Added script to detect extra chars or other problems in casemets
and alto files. Derived from the alto style importer.

	new file:   scripts/compare_alto_case.py
	modified:   scripts/tests/test_helpers.py


ChefAndy:
Forgot to add the script to go through all of the cases and check

PR #214, opened 2018-03-30
Remove separate capapi db

jcushman:
Did I miss anything?

Production deployment process:
- Before deploy, pg_dump the data from `capapi.capapi_apitoken`, `capapi.capapi_apiuser`
- Before deploy, `delete from capdb.django_migrations where app='capapi';`. This will cause the next migrate to re-run the capapi migrations in the default schema.
- Deploy.
- Load the data from `capapi.capapi_apitoken`, `capapi.capapi_apiuser` into the newly-created tables in the `capdb` schema.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,6 @@ the `capstone` virtualenv activated.
 Set up a postgres database:
 
     (capstone)$ psql -c "CREATE DATABASE capstone;"
-    (capstone)$ psql -c "CREATE DATABASE capapi;"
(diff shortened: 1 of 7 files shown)


jcushman:
Fixed tests.

anastasia:
Looks good to me!

bensteinberg (commented):
Looks good. One other step, after deploying to both stage and prod, will be to clean up the config that populates `settings.py`. Before merging this in, I'll run the pg_dump on stage.

PR #216, opened 2018-04-03
API/Test tweaks

jcushman:
A few tweaks while messing with the API tests:

* Add related_name for APIToken
* Avoid catching generic exceptions for login code
* Bump django-extensions version
* Switch auth_client to APIClient instead of RequestsClient
* Drop transaction=True from all tests


anastasia:
looks like attempting to resolve conflicts in github was a mistake:
`test_authenticated_multiple_full_cases`'s `auth_client.get(url)` creates 7 select queries and 1 update, so, line 202 in capapi/tests/test_api.py:
```
    with django_assert_num_queries(select=7, update=1):
        response = auth_client.get(url)
```
I'm a bit trigger-unhappy at the moment and would rather leave it to you to resolve conflicts (probably what I should have done in the first place, sorry if I complicated things).

PR #221, opened 2018-04-03
Update collectstatic settings

jcushman:
- Move `STATIC_ROOT` to `staticfiles/` (location inspired by the [Heroku Django example project](https://github.com/heroku/heroku-django-template/blob/master/project_name/settings.py)). This presumably requires a salt config change.
- Drop compiled css from `static/` source dir.
- Tweak settings so files in `static/` are found by `./manage.py collectstatic` and compiled by `django-pipeline` (also consolidate the PIPELINE settings).
- Add `whitenoise` so Django serves collected static files when `DEBUG=False`. This is handy to have for local testing of django-pipeline, and fine to have and ignore on production. (Ben, just FYI we can also use this to serve the static files directly from Django in production and avoid a separate nginx route for them -- no particular need to switch at this point though.)


anastasia (approved):
Looks great!

bensteinberg (approved):
great!

PR #225, opened 2018-04-06
Fixtures for benchmarking

anastasia:
Attempting to create a bunch of test case data so that speedups of db queries become noticeable.


jcushman:
Cool!

PR #226, opened 2018-04-06
Add bag_reporter fab task

jcushman:
We have a user who would like data for some individual federal reporters. This adds a bag_reporter task to allow that. I also did some refactor to use the pathlib library because I have a curious fascination with that.

### DIFF
--- a/capstone/capdb/tests/test_tasks.py
+++ b/capstone/capdb/tests/test_tasks.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 import pytest
 import bagit
 import zipfile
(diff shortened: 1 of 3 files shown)


bensteinberg (commented):
Looks great, pending import fixes for flake8, and maybe adding a newline at the end of scripts/export.py.

Is `bag.validate()` more idiomatic than `assert bag.is_valid()`, or does it provide better information on failure?

jcushman:
Yeah, `bag.validate()` raises an exception with a description of the problem on failure, instead of just returning `False` -- helpful for figuring out what's wrong if you break the manifest like I did.

PR #227, opened 2018-04-09
Api filter indexing

anastasia:
This should be pretty straightforward: 
- added indexes on fields we're using for filtering
- added indexes on upper case to remove ambiguity on case searches (as per Jay's suggestion)
- also added fuzzy searching (pattern ops) to indexes as per https://www.postgresql.org/docs/9.2/static/indexes-opclass.html
- fixed slugification of normalized_cite

Notes:
- I've added two indexes on a text field, namely `capdb.case_metadata.name`, but I've come across some comments stating that this is a bad idea (it works like a charm locally, but I'm only running it on some 55,000 records). If we're unabashed about adding indexes, then I'm hoping we could chance at these ones and remove them if they are awful (for instance, we can use https://www.geekytidbits.com/performance-tuning-postgres/ 's `Unused Indexes` method to see if these indexes are useless). This might be the place to start investigating something like a solar index. (I was hoping to change this field to a char field of limited length, but as we've found out, the longest entry thus far is 107,757 characters long). Will add a trello card for this.

Followups:
- not currently allowing users to take advantage of LIKE searches, will add a separate pull request for this
- also, filing an issue/question card on trello


jcushman:
Cool, let's give this a try!

PR #230, opened 2018-04-09
API LIKE searches

anastasia:
- allowing users to search on partial names of courts, reporters, cases
- only allowing this for cases right now


anastasia:
per @jcushman — https://stackoverflow.com/questions/1566717/postgresql-like-query-performance-variations/13452528#13452528
`xxx_pattern_ops` only handle left-anchored patterns. Investigating Jay's GIN indexing suggestion

PR #232, opened 2018-04-10
First pass at case save efficiency

ChefAndy:
This trims down the basic update cases pretty well... there might be some more complex cases that could use some trimming down.


jcushman:
Cool, great first pass at this.

PR #234, opened 2018-04-11
API renderers

anastasia:
- add custom renderers for JSON, HTML, XML
- move most of the `get_casebody` logic into our JSON renderer, since so much of it is about checking the requested renderer and body format
- responding with requested format type in case of authorization/authentication errors (similar to what we do right now with with JSON & BrowsableAPIRenderer)

Followup:
- currently I'm keeping the XML and HTML error responses very minimal. In the future, it might be nice to serve up the metadata, though I couldn't think of ways to do that consistent with the structure that our users might be expecting from successful requests.


capstone/capapi/renderers.py
```diff
@@ -0,0 +1,110 @@
...
+
+        if data['casebody']['status'] != 'ok':
+            return """
+                <data>
+                    <error>Casebody Error</error>
+                    <message>%s</message>
+                </data>
+            """ % (data['casebody']['status'])
+        else:
+            case = CaseMetadata.objects.get(id=data['id'])
```
jcushman: Can replace these queries with `data['casebody']['data']`, right?

jcushman (commented):
Just one suggestion ...

capstone/capapi/renderers.py
anastasia: Oh, oops! Totally!

PR #238, opened 2018-04-12
add - removing underline on hovered links

anastasia:
the teeny tiniest of fixes because cap is perfect and nothing else needs to be done


abziegler:
Finally!

PR #240, opened 2018-04-16
Slow query names

jcushman:
So we can treat our slow queries like pets rather than cattle -- quick pass at letting us label slow queries in the admin. Slow query report only shows the label for labelled queries rather than the whole query.

We could also add features like hiding known queries and tracking cumulative time, but I'm not sure it's a big deal at this point.

Also adds ms/query to the report.


bensteinberg (commented):
Wow, that's slick.

Just need to scoot the migration up to depend on 34?

PR #242, opened 2018-04-18
Lazy load jurisdictions to get initial migrate working

jcushman:
This avoids an error when `manage.py migrate` is run on an empty database.


jcushman:
Fixed an unrelated issue that caused tests to fail occasionally based on randomly generated names. Thanks for the debugging help, `pytest-repeat`!

ChefAndy:
Nice catch

PR #244, opened 2018-04-18
Remove default ordering on CaseMetadata

jcushman:
Having a default ordering on CaseMetadata turns out to slow down unrelated parts of the app, like a random `cases = CaseMetadata.objects.filter(reporter=reporter)` in a fab task that results in a slow query. Let's make ordering something you have to opt into.

Is there anywhere else besides the /cases/ endpoint that expects cases to come back in date order?


anastasia:
I don't think so! Awesome.

PR #245, opened 2018-04-18
Let local tests run without collectstatic

jcushman:
Reviving Anastasia & Andy's fix to local tests


anastasia:
Excellent, thanks! Completely flew out of my head.

PR #246, opened 2018-04-18
Staff/user admin permissions

anastasia:
- added Jurisdiction to admin app
- allowing staff users to change jurisdictions but not add or delete (as per https://trello.com/c/GPkEZ5tS/122-add-jurisdictions-to-admin-panel-for-quick-whitelisting-blacklisting)
- added explicit list of permissions in permissions.py
- testing user level permissions

### DIFF
--- a/capstone/capapi/admin.py
+++ b/capstone/capapi/admin.py
@@ -3,7 +3,6 @@ from django.contrib import admin
 
 from . import models
 
-
 def authenticate_user(modeladmin, request, queryset):
     """
     This method will override old key_expires fields by setting it to timezone.now()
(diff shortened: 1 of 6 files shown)


jcushman:
Looks good! This works for now, but note before you invest too much more in this direction that the built-in Django auth system can replace the `staff_level_permissions` stuff you're doing here -- we can define a "staff" auth_group (or whatever other groups we want) via the admin GUI with whatever combination of add/edit model permissions we want, and put selected users in that group. The general Django idea is that `is_staff` users are allowed to log into the admin but can't actually see or do anything there, unless an `is_superuser` user puts them in an auth_group that confers particular permissions.

anastasia:
oh d'oh! Well that completely invalidates this code, django keeps thinking of everything.

PR #247, opened 2018-04-21
Django auth refactor

jcushman:
This turned into a huge refactor. Sorry I didn't break it out more! Summary of changes:

- Model changes:
    - Add email_verified column, default is_active to True, default total_case_allowance to 0. This allows self-registered users to log in far enough to see a prompt to verify their email address, but does not allow them to download blacklisted cases without our review.
    - Rename key_expires to nonce_expires, as suggested in a comment

- Views refactor:
    - Use built-in django views for login, logout, password change, password reset
    - Use django views and forms, rather than DRF views and serializers, for our user-related views (register, verify email, resend verification email, view user details)
    - Split api views, user views, and doc views into separate files
    - Drop custom 404 view -- default Django behavior should work

- DRF configuration changes:
    - Disable query parameter auth. Enable Django session auth instead.
    - Default the API format to JSON, if no Accept: header or format= parameter is provided. Change tests and docs to match. (This came up while adjusting the behavior of check_response in the tests.)
 
- UI changes:
    - Add log in/log out in the top right (TODO: css is messed up for this)
    - Add menu items in the docs TOC (TODO: doesn't really make sense to put these here, just wanted somewhere for them to live for now):
        - Menu items for logged out users: log in, register
        - Menu items for logged in users: log out, change password, view account details
    - Display tweaks: 
        - Lists have bullets by default
        - "caselaw access project" in top left is a link to homepage
 
- Template organization:
    - Move all user-account related templates into templates/registration/, where the default Django views keep their templates
    - Rename docs.html to home.html
 
- Test helpers:
    - `check_response` can now handle either Django or DRF responses, checks based on content_type instead of renderer, and can check for a substring in response contents
    - The `auth_client` fixture returns a user who is authorized both via DRF token header and Django auth


anastasia:
Looks and works flawlessly! 
I want to add another TODO:
We have a few too many angry validation errors when registering as new users, so we need to do template validation sweep 

screen shot 2018-04-23 at 10 39 28 am

Another thing to keep in mind is we will need to delete the tokens of existing, authenticated users and ask them to re-authenticate themselves (so that a new token is created and `email_verified` is set to true).

PR #251, opened 2018-04-25
Add flower for celery monitoring

jcushman:
Just adding the pip requirement at this point. As a first pass, I'm thinking we can log into prod and manually run `flower -A config` to run flower locally on port 5555, then open an ssh tunnel to view it in a browser. If it turns out to be helpful we could have it run all the time.


bensteinberg (approved):
love it

PR #254, opened 2018-04-25
Add recommended Django SSL settings

jcushman:
Will have to test this with our proxy setup, but these are settings recommended by https://docs.djangoproject.com/en/2.0/topics/security/ to redirect requests to https and ensure that session cookies are only set on secure requests.


bensteinberg (approved):
This looks fine; I think it should be a no-op in our environment, as nginx is redirecting port 80 traffic to 443.

PR #256, opened 2018-04-25
Queryset casexmls

anastasia:
fixes https://trello.com/c/qBBebaPK/138-fetch-all-casexml-for-case-endpoint-in-single-query
I think this is potentially a nicer fix? We can use `select_related` for `OneToOne` relationships. I'm seeing 106+ queries, for instance (when getting page_size=100), reduced to 7.


jcushman:
Looks good!

The benefit with using a separate query to fetch the casexmls would be that it's more efficient if a request mostly returns blacklisted cases -- you don't fetch 100 CaseXMLs and throw away 99 of them. But I don't know how common that will be. I'm cool with going with this for now.

PR #258, opened 2018-04-26
smartquotes

anastasia:
fixes https://trello.com/c/TvH0oFbo/154-check-funky-characters-in-api-responses-wrong-response-headers-for-character-encoding

- replace smartquotes with dumb quotes when extracting casebody

Note: this issue only shows up in Safari
There is still the issue of "de- • fendant" in the example case mentioned in the Trello card, but I'm uneasy about removing all bullets — this might be best for a case by case basis (though @ChefAndy confirmed it's definitely an OCR error with this case).

### DIFF
--- a/capstone/scripts/helpers.py
+++ b/capstone/scripts/helpers.py
@@ -1,3 +1,4 @@
+import re
 import shutil
 from lxml import etree
 from pyquery import PyQuery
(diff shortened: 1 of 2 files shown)


ChefAndy:
Regarding your OCR-related aside: we definitely shouldn't touch it— there are bound to be actual bullets in other cases that we wouldn't want to lose. We need to come up with a unified, sensical way to address OCR errors.

anastasia:
After a discussion with @jcushman, opting to close this since this is a Safari-specific bug that should be fixed by Safari.

https://github.com/encode/django-rest-framework/issues/2891

PR #259, opened 2018-04-26
Unlimited access

anastasia:
- allowing datetime setting of `unlimited_access_until` field
- if unlimited access is in effect, not worrying about changing case allowance
- if unlimited access is in effect, showing unlimited access until date + time in user account template


ChefAndy:
saweeeet

PR #260, opened 2018-04-27
Extract and use metadata from VolumeXML

jcushman:
There's two separate things going on in this pull request.

In the first commit, I added a `fetch_relations` function that can efficiently fill in missing data for an instance that didn't use select_related and prefetch_related during the initial query. This turned out to be pretty complicated to get right and I'm not sure it was worth it, but there it is.

I added that to deal with extra queries caused by the second commit, which extracts metadata from volume xml on save and uses it in the API. This makes volumexml/volumemetadata more like casexml/casemetadata.

Here's the story behind that:

Our volume xml has a block at the top like this:

```
        <volume xmlns="http://nrs.harvard.edu/urn-3:HLS.Libr.US_Case_Law.Schema.Volume:v1" barcode="32044057891608" contributinglibrary="Harvard Law School Library">
          <reporter abbreviation="Ill." volnumber="1">Illinois Reports</reporter>
          <nominativereporter abbreviation="" volnumber=""/>
          <voldate>
            <start>1819</start>
            <end>1954</end>
          </voldate>
          <publicationdate>1861</publicationdate>
          <publisher place="Hartford, CT">Press of Case, Lockwood and Company</publisher>
        </volume>
```

We thought those values were all identical to the tracking tool values we have in VolumeMetadata, but it turns out they actually reflect a bunch of corrections made by Innodata and approved by us -- see https://github.com/innodatalabs/rli_query_log/issues/171 for details. In particular they often modified publicationdate, which we have as publication_year. So we want to use the xml values instead.

This PR extracts those into a bunch of VolumeMetadata.xml_ fields, and uses those in the API.

This does extract the reporter information, but does not yet check whether the reporter relationships we have are correct. Saving that part for another day.

Deployment note: once this is deployed we should run `fab update_volume_metadata` to populate the xml_ fields for existing volumes.


bensteinberg (approved):
👍 
I think there's no need to delete any metadata after deployment and before running `fab update_volume_metadata`. Is that correct?

anastasia (approved):
Looks great!

ChefAndy:
This is good. I was considering writing something to get relations for an existing model instance and submitting a PR to Django. Nice feature. 
Also, I'm interested in seeing the discrepancies between the TT data and the Volume XML data now that it's nicely stored in the DB.

ChefAndy (commented):
nice

PR #261, opened 2018-04-30
Jurisdiction slug

anastasia:
Well, this turned out a bit hairier than expected.
- This PR moves to once again using `jurisdiction__slug` instead of `jurisdiction_id` for query_params.
- It redirects from `jurisdiction__name` to `jurisdiction__slug` 

Where it turned out to be a little nuts is in providing an index on `jurisdiction__slug`.
It looks like Django doesn't allow simple index creation on non-primary keys, and so a `to_field='slug'` has been set on `case_metadata`'s `jurisdiction` foreign key.
In order to do this in a way that both forward and backward migrations were allowed to happen seamlessly, three migrations needed to be created.
The reasons for three migrations are:
- Django doesn't allow a simple `to_field` change, so we're creating a duplicate field called `jurisdiction_slug`, copying the content of `jurisdiction` over to it, removing the old `jurisdiction`, and renaming `jurisdiction_slug` to `jurisdiction`.
- Django also complains about both a data and schema changes happening in one migration, and so they had to be split out into three

Here are what each successful migration step should look like (specifically note `jurisdiction` fields):

`0037`

screen shot 2018-04-30 at 12 02 49 am

`0038`

screen shot 2018-04-29 at 9 33 08 pm

`0039`

screen shot 2018-04-29 at 9 33 17 pm

Of particular concern is that now,  `case_metadata.jurisdiction_id` will be actually referring to `case_metadata.jurisdiction.slug`.
`case_metadata.jurisdiction_id` != `case_metadata.jurisdiction.id`

...worth mentioning, maybe, that a much simpler solution to this problem would be changing the jurisdiction slug to be the jurisdiction primary key


capstone/capdb/models.py
```diff
@@ -744,8 +747,8 @@ def create_or_update_metadata(self, update_existing=True, save_self=True):
                     court = Court(**court_kwargs)
                     court.save()
                 case_metadata.court = court
-                if case_metadata.jurisdiction_id and court.jurisdiction_id != case_metadata.jurisdiction_id:
-                    court.jurisdiction_id = case_metadata.jurisdiction_id
+                if case_metadata.jurisdiction and court.jurisdiction != case_metadata.jurisdiction:
+                    court.jurisdiction = case_metadata.jurisdiction
                     court.save()
```
anastasia, line 752: This might be a huge edge case, but it strikes me as something possibly problematic (even before the changes in this pull request). If `case_metadata.jurisdiction_id` does not equal `court.jurisdiction_id` then should we change a court's jurisdiction based on what might be a print error?

jcushman:
After discussing with Anastasia, pulled the first three commits -- left out the commit for now that changes the foreign key to slug. I'm working on something to copy denormalized fields to CaseMetadata, which also solves the indexing issue here.

PR #262, opened 2018-04-30
small style fixes

anastasia:
- changed all non-header fonts to something more easily readable
- fixed header links, removed account related links from api table of contents
- some other small css fixes including some changes to account view

screen shot 2018-05-02 at 1 54 22 pm


bensteinberg:
I like the new font. Can we host it locally? To be discussed today, maybe.

PR #264, opened 2018-05-08
First cut at exposing Swagger/OpenAPI specification

bensteinberg:
https://drf-yasg.readthedocs.io/en/stable/readme.html#quickstart


bensteinberg:
We almost certainly don't need all three routes for consumption by an API gateway.

ChefAndy:
Swaaaaag

PR #266, opened 2018-05-09
Page breaks

ChefAndy:
As usual, this is radically different from the way I started to do it but it's a lot simpler and seems to work well. Had what seemed to be some problems but I was mostly being dumb and they were easy to solve after I worked on other things for a few days. woot.


capstone/scripts/generate_case_html.py
```diff
@@ -41,6 +42,16 @@ def generate_html(case_xml, tag_map=tag_map):
         # remove the namespace from the tag name
         tag = element.tag.split('}')[1]
 
+
+
+        element_text_copy = element.text
+
+        if element_text_copy is None and tag != 'pagebreak':
+            if len(element.getchildren()) == 0:
```
jcushman: Remove debugging line

jcushman (changes requested):
Just one tweak ...

jcushman (approved):

PR #268, opened 2018-05-11
Denormalized fields

jcushman:
As is becoming my habit this is pretty heavyweight for the problem it solves, but I hope sets us up in general to add any API filters we want with good performance and simple design.

Jay keeps telling us: "A handful of INNER JOINS with a handful of predicates and some ORDERS, OFFSETS, and LIMITS aren't ever going to be fast. Denormalize! You've only got 10 million records or so."

In other words, it's hard to tune our /cases/ endpoint when it requires a sql query on CaseMetadata joined to a bunch of other tables, and some kinds of filters are just impossible to run quickly. But it becomes easier if we copy all the values we need into the CaseMetadata table and just query that. And that's not a huge deal when we're talking about only 10 million rows with some redundant data in them.

Difficulty level: there's a reason we generally want databases to be normalized, meaning every fact exists in exactly one place. Denormalized data is (at best) a cache, with all the misery of caches, and at worst becomes two sources of competing truth. Yuck. (Our XML is a massive normalization violation, which is why our XML update code is such a headache ...)

The cleanest way I know to handle this is to populate the denormalized fields (aka cached fields) with database triggers, so they can't really ever go out of sync. So let's say we have `capdb_casemetadata.jurisdiction_id` as a foreign key to `capdb_jurisdiction.id`, and we want to make sure that `capdb_casemetadata.jurisdiction_name` is always a perfect copy of `capdb_jurisdiction.name`. We need two triggers:

* We need a trigger on `capdb_casemetadata` so, when `capdb_casemetadata.jurisdiction_id` is inserted or updated, the value from the related `capdb_jurisdiction` row is fetched and used to set `capdb_casemetadata.jurisdiction_name`.
* We need a trigger on `capdb_jurisdiction` so, when `capdb_jurisdiction.name` is updated, the `capdb_casemetadata.jurisdiction_name` value for all related `capdb_casemetadata` entries is also updated.

So that's the business: for each cached field, make sure that those two triggers will be called when needed, and then you can happily use the cached field as a perfect copy of the related field.

This PR does that. If you set something like this on a model:

```
class CaseMetadata(models.Model):
    denormalized_fields = {
        'jurisdiction_slug': 'jurisdiction__slug',
        'jurisdiction_name': 'jurisdiction__name',
        'jurisdiction_name_long': 'jurisdiction__name_long',
        'jurisdiction_whitelisted': 'jurisdiction__whitelisted',
    }
    jurisdiction_name = models.CharField(blank=True, null=True, max_length=100)
    jurisdiction_name_long = models.CharField(blank=True, null=True, max_length=100)
    jurisdiction_slug = models.CharField(blank=True, null=True, max_length=255)
    jurisdiction_whitelisted = models.NullBooleanField(blank=True, null=True)

    @property
    def denormalized_jurisdiction(self):
        """
            Equivalent to self.jurisdiction, but populated based on denormalized fields.
        """
        return Jurisdiction(
            id=self.jurisdiction_id,
            slug=self.jurisdiction_slug,
            name=self.jurisdiction_name,
            name_long=self.jurisdiction_name_long,
            whitelisted=self.jurisdiction_whitelisted)
```

then the `update_postgres_env` fab task (which we call via `fab migrate`) will make sure that two triggers are set up:

* `capdb_casemetadata` runs `denormalize_dest('capdb_jurisdiction', 'jurisdiction_id', 'id', '{"slug": "jurisdiction_slug", "name": "jurisdiction_name", "name_long": "jurisdiction_name_long", "whitelisted": "jurisdiction_whitelisted"}')` on INSERT or UPDATE
* `capdb_jurisdiction` runs `denormalize_source('capdb_casemetadata', 'jurisdiction_id', 'id', '{"slug": "jurisdiction_slug", "name": "jurisdiction_name", "name_long": "jurisdiction_name_long", "whitelisted": "jurisdiction_whitelisted"}')` on UPDATE

Those two functions are in turn defined in `services/postgres/` (written in Javascript with Postgres's builtin v8 engine, yay!) and use the inputs to keep everything in sync.

It's a long walk to get there, but the upshot is (as long as everything's been saved and fetched from the database), you can use `CaseMetadata.denormalized_jurisdiction` as a free copy of `CaseMetadata.jurisdiction`, and `CaseMetadata.jurisdiction_name` as a copy of `CaseMetadata.jurisdiction.name` and so on. I then used those to tweak the API so /cases/ doesn't join against the jurisdiction table at all, and can use an indexed query to filter on `jurisdiction_slug`.


bensteinberg (commented):
This looks awesome. I think for tests to pass on Travis, we need to do `sudo apt-get update` and `sudo apt-get install postgresql-9.6-plv8` in `before_install` and run `psql -U postgres -c "create extension plv8"` in `before_script` (or maybe `before_install`?)

bensteinberg:
👍

anastasia (approved):
Cool!!

PR #271, opened 2018-05-16
Fix total_sync_with_s3 task

jcushman:
Whoops.


ChefAndy:
I think we should just put a @pytest.mark.skip() on that test for now

ChefAndy:
ha, beat me to it

PR #273, opened 2018-05-17
Delay volume checksums

jcushman:
As discussed yesterday with Andy ...

Updating case xml or page xml has a ripple effect -- it also requires updating volume xml, because volume xml includes md5s of all the case/page xmls.

This means if we update something simple like the court short name for 1000 cases in a volume, we also create 1000 versioned copies of the volume xml, which is a lot of unnecessary storage.

The proposed fix in this PR is to instead set a flag when modifying case/page xml, `VolumeMetadata.xml_checksums_need_update`, which indicates that the volume xml is out of date and should be updated with new checksums before use.

We currently don't use or expose the volume xml in any way, so there's no code that actually checks this flag, but if and when we write code that does use the volume xml, it would then be necessary to call `VolumeXML.update_checksums()` for each volume with `xml_checksums_need_update == True`. I'm not sure if there's a way to make that automatic -- maybe when we get around to needing it something will suggest itself.

While I was changing the `process_updated_xml` case -> page save cascade, I also tweaked a few lines to avoid a redundant `self.page_xmls.all()` query, and added a short-circuit to bail out near the top of the function if `<casebody>` wasn't modified.

*Long-running migration:*

This includes a migration which I expect to take a very long time. It adds a new "size" field to all the XML tables, and then populates that field using a mass update query.


bensteinberg (commented):
Looks good. Is it possible to condition disabling/reenabling the versioning trigger on its presence, or would it be better to set the trigger up at Travis?

jcushman:
Hmm -- there's no "disable trigger if exists" in postgres, and no clean way to make sure the triggers exist. I think the simplest thing is to just have the migration drop the triggers, and then let them be recreated by `set_up_postgres` after migrations are applied. Not a huge deal since `set_up_postgres` drops and recreates the triggers every time anyway.

PR #275, opened 2018-05-25
Cache everything

jcushman:
The goal with this PR is to have Cloudflare cache everything that can be cached, while not caching anything that cannot be cached.

In particular we want to set `Cache-Control: s-maxage=LOTS_OF_SECONDS` as a header on all cacheable responses. `s-maxage` controls reverse proxies, so will instruct Cloudflare, but not browsers, to cache those responses for an hour/day/whatever we think is appropriate.

A cacheable response is one that is intended for a logged-out user, and OK for Cloudflare to play back for other logged-out users, or intended for a logged-in user, and OK for Cloudflare to play back to other logged-in users. (In the latter case, that means it might have some general logged-in-user HTML, but doesn't contain any private information or metered cases.) Cloudflare will take care of showing logged-out cache items to logged-out users and logged-in cache items to logged-in users.

Detecting cacheable responses is handled by middleware, so it should be safe to create additional views from now on and have them automatically cached or not cached. The middleware checks for non-cacheable views, such as a view that uses CSRF, accesses request.user attributes, or sets cookies, and adds the cache header to any view that passes all tests. Checking for request.user access in particular takes care of most cases where we might show logged-in pages to the wrong user.

See `test_cache.py` for a good overview of how this works in practice. If we add or think of any other edge-case views, we should throw them into the list to test.

DEPLOYMENT NOTE: Once this is deployed, we can set up Cloudflare to cache based on upstream headers. I expect that we will then have a bug where Cloudflare caches HTML and JSON responses under the same cache key, because it isn't using the `Accept:` request header as a cache key yet. When we confirm that issue we can ask them to update our custom key.


bensteinberg (approved):
Looks great!

PR #277, opened 2018-05-25
Fix migration

jcushman:
Issue with last migration was that `CREATE OR REPLACE VIEW` doesn't work if you're changing the columns in the replacement view. Weird.

Learning from our last adventure, I also took this opportunity to split the new-column migrations from the update migrations so we can debug the latter more easily on prod if necessary.


bensteinberg (approved):
:+1:

PR #279, opened 2018-05-30
Court name normalization

ChefAndy:
This should be tested on a larger body of data before it's run.


ChefAndy:
i merged in develop and all the tests are dying. Gotta see what's up.

capstone/fabfile.py
```diff
@@ -205,7 +206,7 @@ def add_test_case(*barcodes):
     from django.core import serializers
     from tracking_tool.models import Volumes, Reporters, BookRequests, Pstep, Eventloggers, Hollis, Users
     from capdb.storages import ingest_storage
-    from scripts.helpers import parse_xml, copy_file, resolve_namespace, serialize_xml
+    from scripts.helpers import copy_file, resolve_namespace
```
anastasia: seems like there are two import lines from scripts.helpers
Want to merge them?

anastasia:
Just one comment, otherwise looks good to me!

capstone/fabfile.py
ChefAndy: Nice catch, thanks!

PR #280, opened 2018-05-30
Better errors

ChefAndy:
Some more verbose errors. I've definitely got a much better handle on the internals of this, which is nice. 

Were we looking for more complex logic with the errors? Seems like the more specific the better considering the usage caps, etc.


ChefAndy:
i merged in develop and all the tests are dying. Gotta see what's up.

ChefAndy:
NOw the build is working?

PR #281, opened 2018-06-01
Data viz continued

anastasia:
- updated overview style (showing map now, instead of bar graph)
- added by jurisdiction view
  no longer loading large complicated json, instead using ajax requests to get info per jurisdiction
- fixed sidebar styles across the site
- rewrote tasks + fab task for data viz

to dos in the near future:
- create svgs for jurisdictions like Native American
- figure out what to do for Regional?
- create "Comparison" view, allowing people to compare two jurisdictions?
- add more visualizations! go wild!

to create data, run two fab tasks:
`fab count_data_per_jurisdiction`
and then
`fab count_case_totals`

These (as well as related tasks) are less optimized than before, but hopefully more humane (both to debug and also then to serve to the front-end). We create a json file for each jurisdiction, and then in the second fab task, use those to count up the case totals. 

Still not ecstatic about the styling, but hopefully it's progress. 

overview:
screen shot 2018-06-01 at 3 26 31 pm

and detail view:
screen shot 2018-06-01 at 3 26 57 pm


jcushman:
Sorry for the delay on this -- awesome stuff!

PR #286, opened 2018-06-14
Use partial indexes for cases endpoint

jcushman:
Let's try switching these to partial indexes -- I think it might help with generating counts for API responses.

Result of this pull is to switch to indexes like this on casemetadata:

```
    "capdb_casem_decisio_1d682b_partial" UNIQUE, btree (decision_date, id) WHERE jurisdiction_id IS NOT NULL AND court_id IS NOT NULL AND NOT duplicative
    "capdb_casem_jurisdi_7edf20_partial" UNIQUE, btree (jurisdiction_slug, decision_date, id) WHERE jurisdiction_id IS NOT NULL AND court_id IS NOT NULL AND NOT duplicative
```

I think this might help with our slow count queries that look like this:

```
SELECT COUNT(*) AS "__count" FROM "capdb_casemetadata" WHERE (NOT ("capdb_casemetadata"."duplicative" = ?) AND "capdb_casemetadata"."jurisdiction_id" IS NOT NULL AND "capdb_casemetadata"."court_id" IS NOT NULL AND "capdb_casemetadata"."jurisdiction_slug" = ?);
```

which currently run much faster if you remove the NOT NULL predicates.


bensteinberg (approved):
Looks great

PR #302, opened 2018-06-21
dbs on the move

anastasia:
```
DATABASES = {
    'default': {
        'NAME': 'capapi',
        ....
    },
    'capdb': {
        'NAME': 'capdb',
        ...
    },
    ...
}
```
`capapi` db has users and admin django stuff. It is now our `default` database.
`capdb` db only has cap related data ("only")

capapi is now our default because capapi has `AUTH_USER_MODEL` and django admin stuff has separation anxiety. It needs to be in the same db. 

It turns out, the whole time we've had multiple databases (before we moved to one) pytest was only doing teardowns for the `default`, a.k.a `capstone`. This worked for us then because all our tests had to do with testing cap data, but now that cap data lives in a non-default db, it requires [a hack](https://github.com/harvard-lil/capstone/compare/develop...anastasia:capapi-moves-dbs?expand=1#diff-fe517f3c724f4f39aadaa4d38757fcccR29) . This is the subject of a four-year ongoing discussion on [pytest's github.](https://github.com/pytest-dev/pytest-django/issues/76
)

Things I am unhappy with:
- The db naming. I renamed from capstone because the project is named capstone and default dbs have a general tendency to be named the same things as the projects, so it might get confusing/difficult later on. If you buy that, what should the names be? capapi and capdb? weird. users and cap? users is not the only thing on that db. default and cap? 
- Explicitly specifying the `capdb` database seems lame, but also might be nice for being extra careful with where our data lives. Previously when we were using `django.db.connection` instead of `django.db.connections`, we weren't specifying a db, but instead we were piggy backing off of the hardcoded `default` db somewhere in the depths of django code. That bit us (me) in the butt when moving to a new setup. (Also that variable naming is just about the worst, no? `connections`? You can barely see the difference). 

Assuming neither of the points above (nor any others I haven't thought of) bother you,

### _**In order to merge, many steps must be taken!!!**_

```
$ # back up everything, everything, everything
$ # if making changes on prod db, stop server
$ # <-- if you merge PR and install the code here and make changes on prod db, from here on out prod will no longer work for a while -->
$ psql
$ ALTER DATABASE capstone RENAME TO capdb;
exit psql
$ pg_dump -t auth_group capdb | psql capapi
$ pg_dump -t auth_group_permissions capdb | psql capapi
$ pg_dump -t auth_permission capdb | psql capapi
$ pg_dump -t capapi_capuser capdb | psql capapi
$ pg_dump -t authtoken_token capdb | psql capapi
$ pg_dump -t django_admin_log capdb | psql capapi
$ pg_dump -t django_content_type capdb | psql capapi
$ pg_dump -t django_session capdb | psql capapi
```
[This](https://stackoverflow.com/questions/30545562/django-column-name-of-relation-django-content-type-does-not-exist) _might_ be optional:
```
$ psql capapi
$ ALTER TABLE django_content_type ADD COLUMN name character varying(50) NOT NULL DEFAULT 'someName';
```
But you might find running this (below) that it is not:
```
$ python manage.py migrate --fake-initial
$ python manage.py migrate --database=default
$ python manage.py migrate --database=capdb
```
is everything cool?
```
$ psql capdb
$ DROP TABLE auth_group;
$ DROP TABLE auth_group_permissions;
$ DROP TABLE auth_permission;
$ DROP TABLE capapi_capuser;
$ DROP TABLE authtoken_token;
$ DROP TABLE django_admin_log;
$ DROP TABLE django_content_type;
$ DROP TABLE django_session;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ This repository contains Capstone and CAPAPI, the applications written by the Ha
 - [Documentation](#documentation)
 
 ## Project Background <a id="project-background"></a>
-The Caselaw Access Project is a large-scale digitization project hosted by the Harvard Law School [Library Innovation Lab.](http://lil.law.harvard.edu "LIL Website") Visit https://case.law.harvard.edu for more details:
+The Caselaw Access Project is a large-scale digitization project hosted by the Harvard Law School [Library Innovation Lab.](http://lil.law.harvard.edu "LIL Website") Visit https://case.law.harvard.edu for more details.
 
 ## The Data <a id="the-data"></a>
 1. [Format Documentation and Samples](#documentation-and-samples)
@@ -130,7 +130,8 @@ Capstone should now be running at 127.0.0.1:8000.
 We have initial support for local development via `docker compose`. Docker setup looks like this:
 
     $ docker-compose up &
-    $ docker-compose exec db psql --user=postgres -c "CREATE DATABASE capstone;"
+    $ docker-compose exec db psql --user=postgres -c "CREATE DATABASE capdb;"
+    $ docker-compose exec db psql --user=postgres -c "CREATE DATABASE capapi;"
     $ docker-compose exec web fab init_db
     $ docker-compose exec web fab load_test_data
 
--- a/capstone/capapi/tests/test_api.py
+++ b/capstone/capapi/tests/test_api.py
@@ -271,17 +271,15 @@ def test_authenticated_multiple_full_cases(auth_user, api_url, auth_client, thre
     for extra_case in three_cases[1:]:
         extra_case.jurisdiction = jurisdiction
         extra_case.save()
-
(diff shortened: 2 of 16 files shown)


meangrape:
Should you delete the dumped tables from `capdb` after they get moved to `capapi`?

anastasia:
Nah, leave em.

jcushman:
This looks good to me! Having to specify capdb in `transaction.atomic` is a bummer but I don't have a better idea. Want to update the README as well?

I'll leave to Ben to pull this when he wants to sort out the migration on stage.

bensteinberg:
This looks great slash terrifying. @anastasia, can you point out exactly where in the process above you'd merge the PR/install the new code?

anastasia:
@bensteinberg I added a comment up top. I think the PR merge/new code installation should happen basically at the very beginning. It's just a matter of prod being down for a while if changes are happening directly on the prod db.

PR #309, opened 2018-06-26
Add fab task to compress volumes on S3

jcushman:
This is a first pass at adding the `fab compress_volumes` task and `compress_volume` celery task, which compress volumes from S3 using the CAPTAR format.

The idea is not to run these on prod, but for Ben to deploy them to a new AWS minion using salt.

The new minion will need four extra packages installed: optipng, openjp2, mozjpeg, and ImageMagick. Install steps are documented in `compress-worker.dockerfile`, and the docstring for `fab compress_volumes` explains how to test the dockerfile locally.


bensteinberg (approved):
Wow, that was fast

PR #312, opened 2018-06-29
Upgrade celery

jcushman:


jcushman:
... also add sqs dependencies for celery

bensteinberg (approved):
+1

PR #314, opened 2018-07-02
Create tar in ram before sending to S3

jcushman:


meangrape:
So, assuming you were using all 20 threads, you could consume up to 20GB of RAM before starting to spool to disk?

bensteinberg (approved):
LGTM

jcushman:
@jayed, it's just 1GB per worker -- the threads are used while prepping the individual files to tar, then they all go into one tar file that can get spooled to disk.

If we end up needing to optimize further, something to try might be just keeping everything in RAM and setting up swap to handle the spooling.

PR #359, opened 2018-07-02
Use temp file instead of ram

jcushman:
Something about streaming to S3 from RAM after the `tar.add()` call results in an infinite loop. Avoid by using a regular temp file on disk and re-opening. Performance impact appears to be trivial.


bensteinberg (approved):
YES

PR #365, opened 2018-07-10
add - make filtering by court more friendly

anastasia:
Making getting court by slug consistent. Also added returning court id in court json.


jcushman (changes requested):
This will result in an unindexed query. We currently have an index on `'court_id', 'decision_date', 'id'`. To switch to slug we'd have to denormalize the court fields the same way we denormalized the jurisdiction fields -- check out https://github.com/harvard-lil/capstone/blob/develop/capstone/capdb/models.py#L697 to see how that works. This also requires a migration to populate the denormalized field -- see capdb/migrations/0039_auto_20180521_1511.py 

Not sure how important it is to use slugs for courts, but definitely doable if you think it's worth it.

jcushman:
Nice!

PR #374, opened 2018-07-11
Enforce and report sitewide limits

jcushman:
- Add a new table `capapi.SiteLimits` that stores two limits: user signups today, and blacklisted case downloads today
- Grant 500-case download permission to the first 50 users to confirm their email address per day
- Block blacklisted case downloads for all users after 50,000 sitewide downloads per day
- Email Django admin email address daily with report of download count, user signup count, and user signup emails

Note: I didn't yet add an automatic process to give 500-case download permissions to users from previous days. Not sure if we'll need that, or can just keep raising the ceiling as we get more popular.

*Deployment Notes*

This requires celerybeat to run a task daily to reset the limits and email us.


bensteinberg (approved):
LGTM

PR #386, opened 2018-07-24
Add validate_captar_volumes task

jcushman:
This includes a fun django-storages class that can read individual gzipped files from our indexed .tar files without untarring.


bensteinberg (commented):
Wow, nice!

PR #388, opened 2018-07-27
Don't require volumes to have cases

jcushman:
With this you can rerun `fab validate_captar_volumes` and should get the failures turned to successes (which you can see in the validation/ folder in the S3 bin -- the successes are 6 bytes).


bensteinberg (approved):
+1

PR #389, opened 2018-07-31
Included unredacted volumes in captar

jcushman:
Let's try this with a couple of individual volumes:

```
            fab compress_volumes:32044057891608_redacted
            fab compress_volumes:32044057891608_unredacted,storage_name=private_ingest_storage
```


bensteinberg (approved):
:+1:

PR #404, opened 2018-08-14
Use only alphanumeric characters in normalized cite lookup

jcushman:
Adam noticed that cases were found in h2o for "1 Ill. App. 17" but not "1 Ill.App. 17" because of the spacing difference. That's because slugify encodes those as "1-ill-app-17" and "1-illapp-17" so they don't match.

This PR changes the cite slug function so it strips everything but alphanumeric, encoding the normalized cite as "1illapp17", so cases should be found regardless of spacing/punctuation in the cite lookup.


anastasia:
Excellent. Thanks!

PR #409, opened 2018-08-16
Case styling

jcushman:
Some more polish on the case HTML.

Changes:

* Some general re-thinking of how to handle footnote numbers in different formats:
 * Footnote numbers in text are kept in the text, but we make sure there's a space between the number and the first word, to avoid getting a lot of "1The" in analysis.
 * Footnote numbers are stripped from delivered XML casebody, because they're duplicative of the `label="foo"` attribute.
 * Footnote numbers in HTML are delivered as an `<a>` tag that links back up to the footnote reference in the text.
* Use Django's template system, bootstrap styles, and SCSS, so cases match the rest of the site. (This replaces Bulma -- sorry, Bulma!)
* Apply CSS directly to the semantic classes like ".parties" instead of adding presentation styles, so the contents of `<section class="casebody">` exactly match the HTML from a JSON `body_format=html` request.
* Add a `<section class="head-matter">` element to contain all the stuff before the first opinion.
* Style the labels like "headnote" in the head matter so they're off to the left of the text and a little greyed out.
* Highlight footnotes when clicking a link from the text.
* Add a "full_cite()" attribute to CaseMetadata for use as the title and h1 of the page. Formats the name of the case approximately as it would appear in a legal brief: "<short name>, <cite1>, <cite2>, <citeN> (<year>)".

Here's a screenshot:

image


ChefAndy:
🔥 Much improved!

PR #411, opened 2018-08-17
Break out opinions in casebody text format

jcushman:
As discussed in cap-dev, this changes the "casebody" format for the default (json), html, and xml body_format options.

* Default format takes the single line of text and breaks it out into a dictionary with "head_matter", "judges", "attorneys", "parties", and "opinions".
* "opinions" is removed as a separate dict.
* "judges", "attorneys", "parties" are dropped from the "html" and "xml" formats, where they are redundant.

SO:

Default looks like this:

```
"casebody": {
    "status": "ok",
    "data": {
        "head_matter": "No. 06-278.\nMORSE et al. ...",
        "judges": [
            "Roberts, C. J., delivered the opinion of the Court ..."
        ],
        "attorneys": [
            "Kenneth W. Starr ...",
            "Deputy Solicitor General ...",
            "Douglas K. Mertz ..."
        ],
        "parties": [
            "MORSE et al. v. FREDERICK"
        ]
        "opinions": [
            {"kind": "majority", "author": "Chief Justice Roberts", "text": "Chief Justice Roberts delivered the opinion of the Court. At a school-sanctioned and school-supervised event ..."},
            {"kind": "concurrence", "author": "Justice Breyer,", "text": "..."},
            {"kind": "concurrence", "author": "Justice ...,", "text": "..."},
            {"kind": "concurrence", "author": "Justice ...,", "text": "..."},
            {"kind": "dissent", "author": "Justice ...,", "text": "..."},
        ]
    }
}
```

"xml" looks like this:

```
"casebody": {
    "status": "ok",
    "data": "<mets xmlns:xlink..."
}
```

"html" looks like this:

```
"casebody": {
    "status": "ok",
    "data": "<section class=\"casebody\"..."
}
```

### DIFF
--- a/capstone/capapi/serializers.py
+++ b/capstone/capapi/serializers.py
@@ -193,6 +193,7 @@ class CaseSerializerWithCasebody(CaseAllowanceMixin, CaseSerializer):
                     c = helpers.serialize_xml(casebody_pq)
                     data = re.sub(r"\s{2,}", " ", c.decode())
                 else:
+                    # serialize to json
                     casebody_pq = helpers.extract_casebody(orig_xml)
 
                     # For the plain text output, footnotes should keep their labels in the text, but we want to make sure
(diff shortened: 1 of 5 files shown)


jcushman:
This also changes the value of the "opinions" field on CaseMetadata to `[{"kind": "majority", "author": "Chief Justice Roberts"}]` instead of `{"majority": "Chief Justice Roberts"}`, to reflect that there can be more than one of each kind of opinion. This means we should eventually re-index the metadata, though there's no urgency since that field isn't currently used.

PR #413, opened 2018-08-17
Add fab task to bulk export case JSON by jurisdiction

jcushman:
This is a first pass at a fab task to supplement or replace our current bulk downloads. The new downloads would mirror our API results, so Ill.zip would essentially be the same as saving all results from `/cases/?jurisdiction=ill` into a file. This is intended to be both smaller and easier to load into a program for bulk processing.

Results are saved in gzipped jsonlines format (one JSON object per line, for reading in series), so the file extension should be `.jsonl.gz`.

Results are almost identical to API results, except that all "url" fields are dropped.

We would offer two versions of each download: one with just text, for NLP, and one with XML if you care about styles, pagination, citation markup or the like. So a given jurisdiction would have two exports:

`fab export_jurisdiction_json:Ill.,ill-text.jsonl.gz`
`fab export_jurisdiction_json:Ill.,ill-xml.jsonl.gz,xml`

### DIFF
--- a/capstone/capapi/serializers.py
+++ b/capstone/capapi/serializers.py
@@ -156,10 +156,13 @@ class CaseSerializerWithCasebody(CaseAllowanceMixin, CaseSerializer):
         fields = CaseSerializer.Meta.fields + ('casebody',)
         list_serializer_class = ListSerializerWithCaseAllowance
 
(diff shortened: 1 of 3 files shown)


bensteinberg (approved):
lgtm

PR #420, opened 2018-08-27
New bulk export format, export model, and API endpoint

jcushman:
Sorry for the huge pull -- let me know if I can help with review.

* New bulk export data format that includes two files for each jurisdiction/reporter we export: one that mirrors a case export with body_format=xml, and one with body_format=text.
* New `CaseExport` model that tracks all of our file exports in the database.
* New API endpoint that exposes available exports to API users.
* Bulk downloads page uses `CaseExport` model.

**Deployment:** 

* Files will still be written to bulk-data/ on prod, because of the new `settings_prod.STORAGES['bulk_export_storage']` setting. So you shouldn't have to create any new folders. 
* The existing "public" and "private" subfolders will be ignored, as that metadata is now tracked in the database, so you can keep those around for now and later delete them.
* `fab bag_jurisdiction` and `fab bag_reporter` no longer take an output filename. As a todo, we probably want a way to queue up those jobs via celery.


bensteinberg (commented):
Wow, this is great. 

The Github UI is barking about what look like wrongly-nested double-quotes on line 8 of capstone/capapi/templates/includes/download_list.html, but I see that previously it was just the same on line 6. I guess the template context can tell the difference between inner and outer?

PR #428, opened 2018-08-29
Fab task to export all cases

jcushman:
To help out with exporting all jurisdictions and reporters:

* `fab bag_jurisdiction` and `fab bag_reporter` now trigger celery tasks that run in the background.
* Add `fab bag_all_cases:2018-08-29` to queue celery tasks to export all jurisdictions and reporters that do not have a successful export with date >= `2018-08-29 00:00:00`. You can be as granular as you want with the date, or omit it to export everything.


jcushman:
Note that you'll want to set celery workers to one per CPU or whatever you're happy with.

bensteinberg (approved):
love it

PR #454, opened 2018-09-14
Nav style tweaks

jcushman:
I made some suggested changes -- feel free to reject/suggest further tweaks/take a whack at it.

- Nav animates when sliding in as well as out
- Main content does not slide anymore (I found that there was a janky flash of white underneath the nav when animating the main content, and I don't think it's necessary)
- Page will still scroll while nav is open, and hamburger nav is as high as it wants to be instead of 100% (I found that not all menu items were always reachable otherwise, and it also keeps the page usable if a user can't get to the close button)
- Menu text can't be accidentally selected (which caused confusing highlights if you dragged a bit while clicking)
- Open menu glyph is "^" instead of "–" to keep items from changing width when opening
- Tweak padding and borders for hamburger items
- Accessibility: JS updates "aria-current" and "aria-expanded"
- Make sure current-page highlight color doesn't get overridden by default styles

Before:

image

After:

image

### DIFF
--- a/capstone/static/css/scss/_nav.scss
+++ b/capstone/static/css/scss/_nav.scss
@@ -45,6 +45,7 @@ nav[aria-label="main"] {
   font-weight: $font-weight-normal;
   letter-spacing: 0.1em;
   z-index: 1;
+  user-select: none;
   @include media-breakpoint-up(lg) {
     -webkit-box-shadow: 0 1px 5px $color-violet-gray-dark;
     -moz-box-shadow: 0 1px 5px $color-violet-gray-dark;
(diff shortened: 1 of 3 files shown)


jcushman:
Rebased to fix merge conflict.

capstone/static/css/scss/_nav.scss
```diff
@@ -106,10 +112,11 @@ nav {
 li.nav-item.dropdown {
   > a.nav-link:after {
     content: url('../../img/arrow-expand.svg');
+    display: inline-block;
   }
   &.show {
     > a:after {
-      content: "\02212";
+      transform: rotate(180deg);
```
rebeccacremona, line 119: In addition to removing the motion, this will have a second subtle effect: characters inserted via css like `content: "\02212"` are read by many screen readers:, but images like `content: url('../../img/arrow-expand.svg');` are not. So, this will make the navigation be read "About", "Tools", etc." Instead of "About minus sign", "Tools minus sign" when the dropdowns are open.

jcushman: Ah, cool!

anastasia:
Going to merge but _might_ override some functionality in a little while, depending on how complex this is

PR #458, opened 2018-09-17
Database-related tweaks for Docker installation

rebeccacremona:
- Mounts the `services` directory so that `/services/postgres` directory is available to `fab init_db`. Mounts the whole `services` directory rather than just the `postgres` subdirectory in case other dependencies: can tease out later if it proves helpful
- Customizes postgres to include to plv8 extension. I attempted to install temporal_tables as well, but kept running into issues (the image's python is too old; missing make dependencies, etc.) and Andy suggested it wasn't worth it)

With these tweaks, the docker installation instructions in the README worked perfectly!

### DIFF
--- a/capstone/docker-compose.yml
+++ b/capstone/docker-compose.yml
@@ -1,7 +1,10 @@
 version: '2'
 services:
     db:
-        image: postgres:9.6
+        build:
+          context: ../services/docker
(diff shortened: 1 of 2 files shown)


jcushman:
Thanks!

PR #462, opened 2018-09-17
Some accessibility tweaks

rebeccacremona:
Here are a few suggestions for polishing the user experience for people navigating with keyboards and people using screen readers. No show stoppers, just icing!

Landmarks
-----------
- typo in "main" role prevented it from being recognized as such
- added a label for "main navigation" (could be 'primary', etc.)
- added  "footer navigation"
- added "table of contents navigation"

image
vs
image
...which is available by list, but more interestingly is exposed to users as they tab through and enter landmark regions (e.g. "entering table of contents").

Keyboard navigation
---------------------
- "return to top" link works for keyboard users
- navigation bar dropdown "buttons" behave more like standard html buttons

Gallery page links and images
------------------------------
- make "homepage" vs. "example" language consistent (it's considered a good practice to use the same text each time you link to a single url, so people don't think they are two different pages)
- remove "go to", which is implied for links
- remove alt texts that will never be read (they are overridden by these links' aria-labels)
add github to github links' text, for the benefit of those who aren't informed by the icon

The `title` and `aria-label` attributes are redundant, but some users may genuinely benefit from the visual `title`-on-hover here, so I think it's fine, even if some automated testing may disagree.

See the commit messages to find the matching code.

I'll try and send another PR with a few more nits tomorrow: I thought this PR was becoming a little unmanageable lol.

### DIFF
--- a/capstone/capweb/templates/about.html
+++ b/capstone/capweb/templates/about.html
@@ -31,13 +31,15 @@ About the Caselaw Access Project
       </div>
       <div class="row">
         {# ==============> MENU <============== #}
-        <div class="list-group sidebar-menu">
-          <a class="list-group-item" href="#overview">Overview</a>
-          <a class="list-group-item" href="#data">What data?</a>
-          <a class="list-group-item" href="#usage">Usage &amp; access</a>
-          <a class="list-group-item" href="#press">Press</a>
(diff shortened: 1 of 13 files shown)


jcushman:
Awesome!

PR #468, opened 2018-09-18
Update API doc text

jcushman:
Some general reorganization, text editing, and additions to API docs.

PR #472, opened 2018-09-18
Semantic Headings

rebeccacremona:
This PR tries to introduce semantic heading levels (1 -> {2 -> {3 , 3},  2}  etc.) without affecting the super-snazzy typography of the design.

I tried to do this unobtrusively by replacing the content tag with the matching `.h1`-`.h6` class from bootstrap (tweaked so that the classes know about cap's custom sizing); I can easily imagine that you'd prefer to organize your css completely differently. Hopefully this strategy isn't too offensive for the time-being.

Since (https://github.com/harvard-lil/capstone/pull/468/files), there are now more deeply-nested heading on the tools > api page than were yet explicitly included in the design. I included a guess here (https://github.com/harvard-lil/capstone/pull/472/files). Since before this PR, the largest heading corresponded to .h4, I went down from there. Unlikely I got it right.

Example before:
image

Example after:
image

I'm not sure I got all the pages that are in play:
- landing
- about
- tools > overview
- tools > api
- tools > bulk
- gallery
- 400/403/404/500 (classes hard-coded, rather than applied via scss)
(contact/sign-up/sign-in looked fine; I didn't look at any other registration-related pages)

If I missed any, just let me know!

### DIFF
--- a/capstone/capapi/templates/400.html
+++ b/capstone/capapi/templates/400.html
@@ -2,11 +2,17 @@
 {% load pipeline %}
 
 {% block content %}
-  <div class="container header-margin">
-    <h3>Bad request</h3>
-    <p>
-      Your request resulted in a "400 Bad Request" response. If you see this error repeatedly,
-      please <a href="mailto:info@capapi.org">write to us</a> to let us know what's going wrong.
-    </p>
+  <div class="basic-page">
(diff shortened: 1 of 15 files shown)


anastasia:
Awesome!!

PR #483, opened 2018-09-19
Move api to subdomain

jcushman:
This moves the `/api/v1/` route to `api.example.com/v1/`. Requires various changes for various people:

**For programmers:** Be sure to use functions for generating URLs instead of hardcoding them. Use `capweb.helpers.reverse` (`{% url %}` in templates) to reverse non-api urls. Use `capapi.resources.api_reverse` (`{% api_url %}` in templates) to reverse api urls.

To find the name of URLs to reverse, use `./manage.py show_urls` or `./manage.py show_urls -c API_URLCONF`.

**For local dev:** Add to your hosts file:

    127.0.0.1       case.test
    127.0.0.1       api.case.test

You can then visit the dev server at `case.test:8000`.

**For deployment:**

Prod deployments require three settings changes:

    PARENT_HOST = 'example.com:8000'
    SESSION_COOKIE_DOMAIN = '.example.com'
    ALLOWED_HOSTS = ['example.com', 'api.example.com']

The server can then have either or both of `example.com` and `api.example.com` pointed to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ This repository contains Capstone and CAPAPI, the applications written by the Ha
 - [The Capstone Application](#the-capstone-application)
 - [CAPAPI](#capapi)
 - [Installing Capstone and CAPAPI](#installing-capstone-and-capapi)
+  - [Hosts Setup](#hosts-setup)
   - [Manual Local Setup](#manual-local-setup)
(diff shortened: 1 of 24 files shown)


bensteinberg (commented):
This looks great. Thanks.

PR #485, opened 2018-09-20
Subdomains configurable from settings

jcushman:
This should let you get everything running for stage without fourth-level subdomains:

```
PARENT_HOST = 'capapi.org'
SESSION_COOKIE_DOMAIN = '.capapi.org'
HOSTS["default"]["subdomain"] = 'stage'
HOSTS["api"]["subdomain"] = 'stage-api'
```

### DIFF
--- a/capstone/config/hosts.py
+++ b/capstone/config/hosts.py
@@ -1,7 +1,5 @@
 from django.conf import settings
 from django_hosts import patterns, host
 
-host_patterns = patterns('',
-    host(r'', settings.ROOT_URLCONF, name='default'),
-    host(r'api', 'capapi.api_urls', name='api'),
-)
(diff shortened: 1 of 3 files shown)


bensteinberg (commented):
Thanks, this looks great. Just to confirm, it should be possible to set
```
HOSTS["default"]["subdomain"] = ''
```
?

jcushman:
Whoops, meant to have it default to `HOSTS["default"]["subdomain"] = ''`.

With that default, prod settings are just:

```
PARENT_HOST = 'capapi.org'
SESSION_COOKIE_DOMAIN = '.capapi.org'
```

(This means stage and prod will get confused by each other's login cookies, so we'll have to log back in when switching between them in the same browser. 🤷‍♂️)

PR #488, opened 2018-09-20
Counts on About Page, API docs cleanup

ChefAndy:
Added some text and some spreadsheet downloads with specific counts to the about page. 

Cleaned up API docs. Took out menu nesting under the "using capapi"  area because the whole thing is about "using capapi" and the nesting didn't show up graphically anyway. Reduced some redundancy created by more recent additions and made formatting more consistent.


bensteinberg:
I think the URLs into the API are going to be at `api.case.law`, not `case.law`; should that TSV file be a template?

ChefAndy:
Donesky

bensteinberg:
Ah shoot sorry, I think the URLs will have the form 
```https://api.case.law/v1/cases/?jurisdictions=ala```
instead of
```https://api.case.law/api/v1/cases/?jurisdictions=ala``` -- that's how they look on stage, anyhow.

PR #495, opened 2018-09-21
Add .bullets list style

jcushman:
I have some `<li>` elements in the About text that want normal bullet styling.

My favorite way to do this would be to remove the top-level `ul {padding-left: 0;} li {list-style: none;}`, and adjust list styles case-by-case. But since that's a bigger change, maybe a good compromise is to revert to default styles within `.page-section`?

### DIFF
--- a/capstone/capweb/templates/about.html
+++ b/capstone/capweb/templates/about.html
@@ -96,7 +96,7 @@ About the Caselaw Access Project
                 Scope limits
               </h2>
               <p>CAP does not include:</p>
-              <ul>
+              <ul class="bullets">
                 <li>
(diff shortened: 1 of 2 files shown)


anastasia (commented):
I don't think that's right, because a lot of other sections will get this change which doesn't really work with our design. 
screen shot 2018-09-21 at 10 48 04 am

Might you consider an optional ul.bullets styled with `list-style-type: circle;` instead? I know it seems trivial, but the big black heavy bullets seem a bit much

jcushman:
I can live with `class="bullets"` -- updated.

PR #525, opened 2018-09-27
Handle BOXES files in captar

jcushman:
BOXES files are done, so let's do another test of captar once this is in, and then maybe we're ready to run the whole thing. Note for the test that the BOXES files only appear in post-1923 volumes.


bensteinberg (approved):
ok!

PR #527, opened 2018-09-27
Django reverse() cleanup

jcushman:
Setting `LOGIN_URL = 'capweb.login'` doesn't seem to do the trick. This upgrades `capweb.reverse` to try all possible subdomains, which gets `LOGIN_URL = 'login'` working for the `/swagger` routes.


ChefAndy:
Thanks for fixing this up. I figured saw this was a problem after it was merged.

PR #542, opened 2018-10-03
Ngram performance tweaks

jcushman:
- `fab ngram_jurisdictions` will, by default, only run for jurisdiction-years that haven't yet been indexed, so it's safe to restart the job.
- Use indexed queries for trimming start and end years.
- Use batch bulk_create


bensteinberg (approved):
:+1:

PR #554, opened 2018-10-05
Add maintenance page export function

jcushman:
- Update the maintenance page to use placeholder variables
- Add `fab url_to_js_string` to save the maintenance page to `maintenance.html` as a standalone file (you have to be running the dev server locally) 

Once exported, the file can be copied with `pbcopy < maintenance.html` and pasted into the Cloudflare worker.


jcushman:
(In progress, please hold.)

jcushman:
This is good to go once the tests pass.

PR #571, opened 2018-10-10
Map + Keep in touch

anastasia:


anastasia:
screen shot 2018-10-10 at 6 45 42 pm
also added a keep in touch section it's pretty empty and kinda lame... what do you think?

PR #573, opened 2018-10-11
Template refactoring and other frontend tweaks

jcushman:
The biggest change here is to move boilerplate HTML from all the templates into `layouts/`, deleting about 500 redundant lines yay.

`layouts` has four templates. We have `layouts/base.html`, which shows the core title/explainer/sidebar/main-content sections, and then `layouts/full.html` for pages like About and API; `layouts/narrow.html` for pages like Login and Contact; and `layouts/wide.html` for pages like Wordclouds and Limericks. There are a number of configuration blocks in `layouts/base.html` (and we can add more) to tweak the display of each page. The `{% block title %}` block by default gets used for the visible title as well as the page title, though you can use `{% block page_title %}` to override.

Other features included in separate commits:

* Footers stick to bottom of short pages, using flexbox
* Sidebars aren't set to 100% of window height, because that pushes the footer really far down on short pages
* For debugging error templates, add local routes like /404.html to see the 404 page
* Contact page now forwards to /contact-success/ on success, so refreshing doesn't resend the email


anastasia:
Note: Tests are failing because of the pytest `--fail-on-template-vars` flag, and the {{ title }} is not always defined in main_base.html template

anastasia:
💯 Love the full/narrow/wide templates and the template reorg in general

Issue #581, opened 2018-10-11
Make CAP sharing friendly


anastasia:
Includes #330

This is what facebook sharing looks like right now:
<img width="527" alt="screen shot 2018-10-11 at 2 45 12 pm" src="https://user-images.githubusercontent.com/1494102/46826569-51f16b80-cd64-11e8-8240-09af655b7ff3.png">

twitter sharing preview is nonexistent. Would be cool if we could make CAP a sharing-friendly website for social media like twitter, facebook, slack.

jcushman:
These are looking good!

PR #590, opened 2018-10-12
Social Media Meta Tags

ChefAndy:
Addresses #581 

- put the meta tags in main_base.html in capweb and api.html and in capapi
- Set up capweb views to pass the correct information
- overrode BrowsableAPIRenderer.get_context() to include appropriate, resource-specific values for title and description
- took basic screenshots for most pages as og_image or twitter card images


ChefAndy:
Hold off on this. Realized that fullurl's fullstatic on API doesn't work.

PR #597, opened 2018-10-12
Social tags

ChefAndy:
- put the meta tags in main_base.html in capweb and api.html and in capapi
- Set up capweb views to pass the image, title and description information
- overrode BrowsableAPIRenderer.get_context() to include appropriate, resource-specific values for title and description
- Added title and description to HTMLrenderer's context
- made a template tag to get an absolute static URL
- took basic screenshots for most pages as og_image or twitter card images


capstone/capweb/views.py
```diff
@@ -58,16 +65,29 @@ def contact(request):
 
     return render(request, 'contact.html', {
         "form": form,
-        "email": settings.DEFAULT_FROM_EMAIL
+        "email": settings.DEFAULT_FROM_EMAIL,
+        'page_image': 'img/og_image/contact.png',
+        'page_title': 'Contact Caselaw Access Project',
+        'page_description': 'Email us at info@example.com or fill out this form. '
```
jcushman, line 71: I think we want `settings.DEFAULT_FROM_EMAIL` instead of `info@example.com` here ...

PR #600, opened 2018-10-15
Fix label for Missouri

jcushman:


jcushman:
(This failed because #599 hasn't been merged yet, but should be fine to pull anyway.)

PR #605, opened 2018-10-15
Update jurisdictions for American Tribal Law Reporter

jcushman:


jcushman:
Hmm ... some sort of Travis error. Replacing my totally wrong code with correct code didn't seem to work. I'm going to try replacing the correct code with even more correct code.

PR #620, opened 2018-10-17
Add tsv column back to CaseText; use rum-friendly result ordering

jcushman:


jcushman:
I'll just kick this off so it can run overnight ...

PR #661, opened 2018-10-30
fix - washington slug in map

anastasia:


jcushman:
Now conflicted by witchcraft. :)

PR #671, opened 2018-10-30
Design tweaks

anastasia:
lots of little fixes including weird title breaks on small screens (cc @abziegler)


anastasia:
screen shot 2018-10-30 at 4 28 08 pm

PR #688, opened 2018-11-02
Update to Ubuntu 16.04 (xenial)

arichiv:
Poking around to see what can be done :-)


jcushman:
Hi, Ari! Thanks for poking around. Right now our production servers are Debian Stretch/Python 3.5/Postgres 9.6, so we're getting as close to that as we can with Docker and Travis. (Docker doesn't have an official `python:3.5-stretch` so we fudge that one.) Definitely on the docket to upgrade production to Py3.7/Postgres 11.

arichiv:
Got it, thanks!

PR #689, opened 2018-11-02
Switch to python3.5-stretch

arichiv:
It looks like there is a 3.5-stretch now! https://github.com/docker-library/python/tree/master/3.5/stretch

In https://github.com/harvard-lil/capstone/pull/688 you mentioned a desire to keep versions as close to lockstep as possible


jcushman:
Thanks, looks like they added this in July!

PR #709, opened 2018-11-13
[WIP] Citation Extraction

arichiv:
#746 

- [x] Support basic citation extraction (420 U.S. 1776)
- [x] Normalize reporter
- [x] Optimize (too many string operations)
- [x] Support Id. citation extraction (Id.)
- [ ] Support exact page extraction (... at 666)
- [ ] Support signal and index extraction (see also ...; cf. ...)


arichiv:
@jcushman how does this sound as an addition to the `CaseXML` object?
```
{
  'name': "$",
  ...
  // Citations by other cases
  'cited_by': [
    {
      'url': 'https://api.case.law/v1/cases/$/',
      'pages': [$]
    }
  ],
  // Citations this case makes
  'does_cite': [
    {
      'url': 'https://api.case.law/v1/cases/$/',
      'pages': [$]
    }
  ]
}
```

`cited_by` and `does_cite` would be powered by a single table of citations indexed to allow lookup either direction.

arichiv:
If we wanted to filter based on what pages of a case were cited (as opposed to what pages citations were made on) that could be done by replacing the pages list with a map of citer page to citee page(s).

arichiv:
@jcushman: I have some scaffolding to get the info into the DB/API! Time to play with the parser (starting with https://github.com/freelawproject/courtlistener/blob/master/cl/citations/match_citations.py)

jcushman:
Hi, Ari. Excited to see this coming along!

A couple of thoughts about API format:

- I'd like to move inbound and outbound under one dictionary.
- In the long run (not necessarily first pass) I'd like to include various details like the cite text, signal, and parenthetical. So I want to design the structure to make that workable if and when we get to it.

So we could end up with something like this (totally open to rephrasing/restructuring!):

```
  {
    "name": "$",
    ...
    "citations": {
      // Citations this case makes
      "outbound": [
        // citation object created from "but see United States v. Legault, 323 F. Supp. 2d 217, 219 (D. Mass. 2004) (Jonas, J., dissenting) (noting historical examples)", cited by the third opinion in the source case, which is a concurrence:
        {
          "url": "https://api.case.law/v1/cases/$/",
          "opinion_type": "concurrence",
          "opinion_index": 2,
          "from_page": $,
          "to_page": 219,
          "text": "United States v. Legault, 323 F. Supp. 2d 217 (D. Mass. 2004)",
          "signal": "but see",
          "author_parenthetical": "Jonas, J., dissenting",
          "parenthetical": "noting historical examples",
        }
      ],
      // Citations by other cases
      "inbound": [
        ...
      ]
    }
  }
```

Buuut this could be a ton of data if there were dozens of outbound cites per case, and potentially thousands of inbound cites. We could have a minimal format as well, with just URLs or IDs for people who just care about the graph:

```
  {
    "name": "$",
    ...
    "citations": {
      "outbound": [$, $, $],
      "inbound": [$, $, $]
    }
  }
```

And then maybe we let you choose the level of detail with a query parameter, like https://api.case.law/v1/cases/?citations=[none/short/full]

As far as sequencing work, I'm a big fan of landing something sooner rather than later, so you could look at just supporting the "short" form first, and just build in a way that will make the long form possible in the future. Also happy to chat if that's not feeling like enough clarity.

Thanks so much for working on this!

arichiv:
Thanks! I like the idea of supporting multiple forms, and can aim for a quick implementation of the short form. To be clear, are you suggesting I use the existing citation field? It has a slightly different purpose (more related to where it was reported), and is a list (not a dict). Should there be a new field (external_citations) or should the citations field be re-worked to reduce confusion?

@jcushman

jcushman:
Oh, good point, I missed the naming collision. How about `"citation_graph"` for the new field?

arichiv:
Sounds good!

arichiv:
API Changes
```
CASE = {
  ...
  'citation_graph': {
    'incoming': [
      {
        'url': $,
      },
    ],
    'outgoing': [
      {
        'url': $,
      },
    ],
  ],
}
```

SQL Changes
```
CitationGraph
  id = AUTO_INC
  src_case = FK/Case
  dst_case = FK/Case
```

arichiv:
* It looks like sub-moduling caselistener won't work due to the breadth of its dependencies (data and code)
* The critical code needed is https://github.com/freelawproject/courtlistener/blob/master/cl/citations/find_citations.py#L509
* This isn't super involved (tokenizing then searching the text), and I think replicating it (with attribution) is preferable to avoid importing a dependency not designed to be importable

awagner-mainz:
Hi all. Please allow me to apologize for lurking on the repo so far. I am very interested in the project and the API and am very happy that development is continuing. Congrats for all of that.

With regards to the issue at hand, I wonder if it would make sense to separate the citation information from the "core" case information and host it at some place like, say, https://api.case.law/v1/cases/435800/citation?direction=(in|out|both)

Is there still room for discussing this?

arichiv:
The current approach was to embed in the case object (your concern) with a query-flag for no/brief/full information. Your question made me wonder if this approach would fall down in cases where pagination is required, and if you are correct to request a different endpoint. What about:
```
https://api.case.law/v1/citation_graph/$CASE$?direction=(in|out|both)&form=(long|short)
```
Then the results can be a simple list (as you suggested), with a direction flag inside each citation instead of on the outer dictionary @awagner-mainz @jcushman

arichiv:
This approach also seems better fitted to bulk API requests as nesting citations inside cases would lead to a lot of duplication.

arichiv:
I should note that although this question must be answered before public release, it isn't blocking me at the moment since I'm focused on loading the data :-)

awagner-mainz:
FWIW, I was thinking of github's API that provides "sub-endpoints" like "https://api.github.com/users/awagner-mainz/following{/other_user}" besides/below the actual "https://api.github.com/users/awagner-mainz" endpoint.

arichiv:
Oh, I like that! Would be down for whatever approach as long as it fits the overall API design

jcushman:
Welcome, @awagner-mainz ! Thanks for jumping in. 

There's definitely still room for playing with the format! I'm also generally fine with the idea of nested endpoints like `https://api.case.law/v1/cases/$CASE$/some_detail` in cases where that's more intuitive.

One usecase where I think the `?citations=[none/short/full]` approach is useful is when a user wants to build a cite graph based on a case search. For example, suppose we're writing a program to build a graph of internal links among Arkansas Supreme Court cases before 1850. Those cases are currently selected like this:

https://api.case.law/v1/cases/?court=ark&decision_date_max=1850-01-01

With a separate citations-per-case endpoint, you would need 1100 queries to fetch the citation information for each case returned by that query. By contrast, if you can just tack on a request to include cite graph information like this:

https://api.case.law/v1/cases/?court=ark&decision_date_max=1850-01-01&citations=short

... you'll only need 11 pagination queries to grab all the information you need. And of course the same param would work for single-case queries, so this would also work if you did need only one case:

https://api.case.law/v1/cases/$CASE$/?citations=short

For this usecase, the separate `/v1/citation_graph/` endpoint would potentially work if if supported all of the same search filters as `/v1/cases/` -- that feels conceptually redundant to me, though, unless it's bringing a lot of additional value.

Totally open to more thoughts -- that's just my first take.

jcushman:
@arichiv, re:

> This isn't super involved (tokenizing then searching the text), and I think replicating it (with attribution) is preferable to avoid importing a dependency not designed to be importable

Makes sense! That was my initial hunch as well. Keep in mind that the copied code may want to become a separate python package at some point, to allow for collaboration with the original authors and other folks interested in cite extraction. For now maybe that just means keeping it in a distinct file/directory with a clean interface to keep it easy to extract later ...

arichiv:
@jcushman Just to check my understanding, it looks like CaseXML objects are ingested (from s3) into your capstone's DB, then serves API requests via django. Each case should expand into 100+ potential citations (especially when `Id.`s are expanded), which can be filtered against a list of known and unknown cases (with some alert around the % of citations we expect to be unresolvable).

1. ~~Are there any demo CaseXML objects with the full case text? https://github.com/harvard-lil/capstone/tree/develop/capstone/test_data/from_vendor doesn't seem to have any.~~ Correction: I see the casebody section in https://github.com/harvard-lil/capstone/blob/develop/capstone/test_data/from_vendor/32044057892259_redacted/casemets/32044057892259_redacted_CASEMETS_0001.xml
2. ~~Is there a reason the casebody isn't shared in the API currently?~~ I keep asking questions too early https://api.case.law/v1/cases/863713/?full_case=true
3. Do we need to require auth for the citation graph (as it's sourced from a casebody which may require it)?
4. The most comprehensive casebody in a test file I found was [In re the Marriage of Lyle](https://github.com/harvard-lil/capstone/blob/develop/capstone/test_data/from_vendor/WnApp_199_redacted/casemets/WnApp_199_redacted_CASEMETS_0036.xml) as it actually had an `Id.` citation to expand. If there's a better Target (or more test data could be integrated into the repo) please let me know, otherwise I'll use the raw text of https://supreme.justia.com/cases/federal/us/532/661/#tab-opinion-1960917 to augment my test data set

jcushman:
@arichiv:

1. You found the case data, but just in case you haven't hit it yet, definitely check out `fab load_test_data` which creates the CaseXML objects in the DB.
3. Let's go with including cite data without auth for now -- certainly I think `citations=short` should work that way.
4. We can add a few more sample cases for sure -- I've been wary of adding too many as far as growing the size of the repo or slowing down tests, but not sure if it's really a problem. That has to happen on our end -- let us know if you have a few cases you'd like to add, based on the live API.

"Each case should expand into 100+ potential citations" -- we have 6.7 million cases, so that's order of half a billion records. I imagine we'll hit some Postgres issues as we go along ...

arichiv:
1 BILLION CITATIONS! @jcushman 

Is the postgress DB sharded? I could imagine using the source case as the sharding hash input and maintaining an index on the destination case

jcushman:
We're not doing any sharding yet, but I think we're going to need it for ngrams -- my naive implementation for ngrams was a nonstarter. Sharding has apparently improved a lot between PG 9.6 and 11, so we might want to mix in a database upgrade somewhere along here as well.

arichiv:
Wanted to provide an update! I'm continuing development of this part in an [external library](https://github.com/AriChivukula/sitation) to move faster and enable other use cases :-)

It has a pluggable [map](https://github.com/AriChivukula/sitation/blob/master/source/mappers.ts#L49)/[reduce](https://github.com/AriChivukula/sitation/blob/master/source/reducers.ts#L40) setup that should allow completion of page and signal detection shortly :-)

mlissner:
This is a bit of a drive-by comment and on a closed bug no less, so apologies for all that, but are you guys storing all the ibid and supras as rows in the DB too? I'm looking at that half a billion rows problem and thinking you could do it more like what we're planning in CourtListener. Our new data model that's going in...one of these days...will just store the following:

> A cites B 22 times

Or whatever. In other words, rather than storing all those, we'll just have a count.

mlissner:
Oh, and sorry, one other thing that caught my eye in my pass through town. We use `cited_by` and `cites` as our vocabulary. Looks like you're using `cited_by` and `does_cite`, which seems a bit odd to my ear. 

And also worth mentioning: Be *super* careful anywhere you put the `cited_by` values. Our top-cited opinion has been cited about 80k times so far. You probably don't want that showing up in HTML anywhere or JSON or...just about anywhere. Even pulling this from the DB is a pain, and even just showing it as a count requires that we keep the count denormalized somewhere as we find citations to it.

Hope this helps!

PR #712, opened 2018-11-21
Use node for compiling javascript

jcushman:
This PR uses node to compile our javascript, and keeps compiled assets in static/dist/.

See the README for how to use this in dev -- basically **if** you want to edit javascript, you have to (1) `npm install` (2) run `npm run serve` in the background while developing (3) run `npm run build` before committing.

Unfinished stuff: we could compile CSS through node as well and get live reloading and versioned CSS libraries and postcss browser compatibility and stuff. Not sure if worth the hassle. I stuck with just JS for now.

Deployment note: this doesn't require anything special for deployment, as compiled assets are in source control.


bensteinberg:
@jcushman -- in deployment, we should still be running `manage.py collectstatic`, I think?

jcushman:
Right, keep running `manage.py collectstatic`

PR #720, opened 2018-11-28
Support new barcode format

ChefAndy:


jcushman:
Nice! I like how clean the logic is for creating missing VolumeMetadata entries. I sent a PR for some tweaks to the filename parsing code at https://github.com/ChefAndy/capstone/pull/1

PR #722, opened 2018-11-28
Data Snippets

ChefAndy:
This is just the back-end. The update_snippets fab task could be run periodically. Weekly? When more tasks are added it probably makes sense to parallelize them.


capstone/fabfile.py
```diff
@@ -804,3 +804,106 @@ def report_multiple_jurisdictions(out_path="court_jurisdictions.csv"):
...
+        "del": { "label": "Delaware", "map_id": "US-DE"},
+        "mo": { "label": "Missouri", "map_id": "US-MO"},
+        "haw": { "label": "Hawaii", "map_id": "US-HI"},
+        "nm": { "label": "New Mexico", "map_id": "US-NM"},
+        "wash": { "label": "Washington", "map_id": "US-WA"},
+        "va": { "label": "Virginia", "map_id": "US-VA"}
+    }
+
+    tally = {}
+    for volume in VolumeMetadata.objects.prefetch_related("case_metadatas__jurisdiction", 'volume_xml__page_xmls').all():
```
jcushman: I think this loop isn't going to work for a couple of reasons. The `prefetch_related` is going to try to pull hundreds of gigabytes of case metadata and page xml into RAM, which will fail. The inner loop over cases also doesn't get the right answers -- for example 'volume_count' ends up as the number of cases, and 'case_count' ends up as the number of cases squared.

I think instead we want to do this as an aggregate sql query, or a few of them, on the capdb_jurisdiction table. Could either use a django Aggregate query (which are always frustrating) or a raw sql call. Happy to chat about this if it's helpful ...

jcushman (dismissed):
Nice! I like the Django view + templatetag approach. There's some issues with the `update_map_numbers` task, which isn't central to the Snippets work -- let me know if you want to fix the task, or mark it as a work in progress and pull as-is.

capstone/fabfile.py
ChefAndy: Ah, good point. I'll change that.

jcushman:
Oh, don't forget to rebase and recreate the migration, since it conflicts with your other PR that I pulled in.

ChefAndy:
already done

ChefAndy:
updated

PR #737, opened 2018-12-13
Add sample_captar_images task

jcushman:
`fab sample_captar_images` will extract the 100th jpg image from each captar volume and store it in captar_storage/samples/ for review.

This doesn't start any celery tasks, just runs serially. It runs in random order and can be resumed if stopped, but will eventually grab an image from all volumes. Run from AWS because bandwidth is excessive -- it fetches more than it has to to extract the target image.


bensteinberg (approved):
Looks great!

Issue #730, opened 2018-12-06
Support range requests for bulk download files


jcushman:
I'm not sure if we currently support the HTTP Range request header, but if not, we should -- it would allow resuming of large bulk file downloads that are interrupted.

Implementation suggestions: https://stackoverflow.com/questions/14324250/byte-ranges-in-django

PR #739, opened 2018-12-17
Add range request middleware

bensteinberg:


bensteinberg:
Addresses #730 -- note that there are no tests for this yet.

bensteinberg:
WIP, do not merge. TODO:
- use regexes
- handle or ignore IF_RANGE
- confirm that `os.fstat` is the right way to get at the size
- clamp start and end correctly
- add tests

bensteinberg:
Also TODO, per @jcushman :
- add `Accept-Ranges: bytes` to 200 responses for files

capstone/capapi/middleware.py
```diff
@@ -92,3 +93,99 @@ def process_request(self, request):
...
+        response = self.get_response(request)
+
+        # is this request eligible for a range request?
+        if response.status_code != 200 or not hasattr(response, 'file_to_stream'):
+            return response
+
+        # parse the range(s) -- at the moment, we're only responding to the
+        # first range, but we coalesce adjacent and overlapping ranges below
+        http_range = request.META.get('HTTP_RANGE')
+        units = "bytes="
```
jcushman: For this block, we can re-use some existing code from Werkzeug:

```
>>> from werkzeug.http import parse_range_header
>>> ranges = parse_range_header('bytes=1-2')
>>> ranges.ranges
[(1, 3)]
>>> ranges.units
'bytes'
```

Note this behaves a little differently in that it returns None for malformed headers -- probably fine in practice.

capstone/capapi/middleware.py
```diff
@@ -92,3 +93,99 @@ def process_request(self, request):
...
+        # multipart response...
+        if len(ranges) > 0:
+            (start, end) = ranges[0]
+
+            # if the range encompasses the whole response, return 200 --
+            if start == 0 and end >= (response_size - 1):
+                return response
+
+            # iterator for range of file
+            def fchunks(start, end):
```
jcushman, line 142: This should use `seek` and much larger chunk sizes for efficiency. Something along the lines of:

```
def fchunks(start, end):
    remaining = end - start
    f.seek(start)
    while remaining > 0:
        chunk = f.read(min(remaining, 2**20))
        if not chunk:
            break
        yield chunk
        remaining -= len(chunk)
```

Though that may have an off-by-one error -- not sure I used start and end correctly.

jcushman (changes requested):

jcushman (approved):

PR #741, opened 2018-12-18
Get tests passing in Docker env

rebeccacremona:
- installs node and npm packages
- installs a local redis on the web/worker image for use by pytest-redis, which cannot be configured to use a remote redis at this time.....
- uses named volumes and the `:delegated` to smooth heavy I/O. Might not be necessary for the local redis, but so far as I can tell harmless
- does not `ADD` code to the image, since up-to-date code is supplied via mounted volumes

Note that tests pass even when the worker containers crash. (Perma doesn't test it's celery workers either)

### DIFF
--- a/capstone/Dockerfile
+++ b/capstone/Dockerfile
@@ -1,11 +1,24 @@
 FROM python:3.5-stretch
 ENV PYTHONUNBUFFERED 1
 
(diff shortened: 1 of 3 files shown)


bensteinberg (approved):
looks great, thanks!

PR #742, opened 2018-12-18
Correctly handle node_modules in Docker setup

rebeccacremona:
My string of incomplete PRs continues.....

Last PR only worked due to an oversight: I had the node_modules dir locally.

This PR, I am 99.999%, properly keeps node_modules inside the image/containers.


bensteinberg (approved):
lgtm, but I thought so last time.... ;)

PR #744, opened 2018-12-27
Update packages that can; note ones that can't

arichiv:
#745 

- [x] Delete and re-build requirements.txt
- [x] Note version incompatibilities in requirements.in


arichiv:
@jcushman ready for review!

bensteinberg:
Thanks, @arichiv! Let me make two suggestions: first, some of your changes are just to the common indentation of comments in requirements.in, like changing line 4, from `tqdm                # script progress bars` to  `tqdm # script progress bars` (and of course GH doesn't show the extra spaces after `tqdm` in the first case) -- this makes it harder to read the whole file, when the comments don't line up, so you might leave them the way they were. I suspect that might be an artifact of whatever editor you're using -- from the frequency of commits, maybe you're using the GitHub UI?

Second, for those using docker containers for development, a change to requirements needs a matching bump to the image versions in capstone/docker-compose.yml, so the containers will be rebuilt accordingly. That's the `image` line in the `worker` and `web` sections. Let me know if you have any questions.

arichiv:
@jcushman I am using the GitHub UI, and changed back the spacing to the old convention :-)

Updated version as well!

bensteinberg:
@arichiv, @bensteinberg here. I just restored some of the whitespace and capitalization in requirements.in to the way they used to be; the point here is to reduce the number of changed lines to the minimum necessary. This makes it much easier to read a PR. By the same token, I recommend you use an editor, and then git on the command line, or an application like sourcetree, rather than the GitHub UI, which creates many tiny commits with non-useful commit messages -- and sends each of us email for each one, for what it's worth! The GitHub UI is very useful for tiny changes, or to changes in text only. Happy to discuss workflow/tooling if you like.

I also note that tests are currently failing, since the kombu developer pulled and rereleased the package, and the version in requirements.txt no longer exists -- you should run pip-compile again.

Finally, this is a lot of changes to requirements for a single commit; we tend to upgrade packages more incrementally. Our test suite is OK, but certainly not complete, so we can't rely on it to protect us from subtle changes in requirements. For package updates, I recommend confirming that our test suite tests the relevant feature, testing the relevant functionality in your local environment by hand, to the extent possible, or writing the appropriate tests! ;)

arichiv:
@bensteinberg thanks for the spacing changes! I'm a big fan of the GitHub UI because I'm a big fan of Travis CI as the check on any changes made. I have emails turned off for what it's worth :-)

The primary change here is re-building requirements.txt from requirements.in, any non-spacing changes are due to newer versions of packages requiring other changes in the application to support them

arichiv:
Tests passed! I'll check off packages below (directly required and updated) as I find tests that cover them:
- [x] bagit 1.6.4 -> 1.7.0 capstone/capdb/tests/test_tasks.py
- [x] beautifulsoup4 4.6.0 -> 4.7.0 capstone/capweb/tests/test_ui.py
- [x] boto3 1.7.48 -> 1.7.84 The latest version of this package breaks tests; I assume this means it is tested
- [x] celery 4.2.0 -> 4.2.1 Assuming minor change
- [x] django-bootstrap4 0.0.6 -> 0.0.7 Looks like a minor change https://django-bootstrap4.readthedocs.io/en/latest/history.html#id1
- [x] django-bulk-update 2.1.0 -> 2.2.0 Appears unused https://github.com/harvard-lil/capstone/search?q=django_bulk_update&unscoped_q=django_bulk_update, @bensteinberg can this be removed?
- [x] django-extensions 2.0.6 -> 2.1.4 Looks like a minor change https://github.com/django-extensions/django-extensions/blob/master/CHANGELOG.md
- [x] django-partial-index 0.4.0 -> 0.5.2 capstone/capdb/migrations/0042_auto_20180614_1524.py
- [x] django-redis 4.9.0 -> 4.10.0 Looks like this is used in prod only, but changes are minor https://github.com/niwinz/django-redis/blob/master/CHANGES.txt
- [x] django-simple-history 2.1.1 ->  2.6.0 Core package: capstone/config/settings/settings_base.py
- [x] djangorestframework-filters COMMIT -> 1.0.0.dev0 Core package: capstone/config/settings/settings_base.py
- [x] djangorestframework 3.8.2 -> 3.9.0 capstone/capapi/tests/test_user_views.py
- [x] drf-yasg 1.7.1 -> 1.12.1 Core package: capstone/config/settings/settings_base.py
- [x] flex 6.13.1 -> 6.13.2 Assuming minor change
- [x] img2pdf 0.3.0 -> 0.3.2 Assuming minor change
- [x] lxml 3.7.3 -> 4.2.6 Core package: capstone/capdb/models.py
- [x] mysqlclient 1.3.10 -> 1.3.14 Assuming minor change
- [x] nltk 3.3.0 -> 3.4 (via ngrams script) capstone/test_data/test_fixtures/fixtures.py
- [x] pip-tools 2.0.1 -> 3.2.0 Assuming system tools are solid
- [x] psycopg2 2.7.1 -> 2.7.6.1 Assuming minor change
- [x] pycurl 7.43.0 -> 7.43.0.2 Assuming minor change
- [x] pytest* ASSUMED GOOD  
- [x] redis 2.10.6 -> 3.0.1 capstone/test_data/test_fixtures/fixtures.py
- [x] reportlab 3.5.2 -> 3.5.12 Assuming minor change
- [x] tqdm 4.11.2 -> 4.28.1 capstone/scripts/tests/test_tt_ingest.py
- [x] whitenoise 3.3.1 -> 4.1.2 Core package: capstone/config/settings/settings_base.py
- [x] webpage2html COMMIT -> 0.3.6 (via fabfile) capstone/capdb/tests/test_tasks.py

arichiv:
ping @bensteinberg

jcushman:
Hi, @arichiv! Thanks for all your work on this!

I wanted to let you know that we hesitated to pull this one in, and almost rejected it. In general our policy is to update packages only as needed, because of a new feature, bug fix, end-of-life policy, or security hole. We do this for a few reasons: (a) it takes a while for the package ecosystem to resolve incompatibilities between new versions of packages; (b) new bugs are introduced at about the same rate they are fixed; and (c) new bugs are easier to fix if we know which parts of our application have been upgraded recently, instead of bumping 50 dependencies at once. For a team of our size supporting the range of software we do, we need to target the most common combination of our dependencies, not the newest.

We did pull this one in because we appreciate your efforts and didn't want to throw out the work you did in testing package upgrades. But I would strongly encourage you in the future to check in before starting efforts of this size -- we'll be more likely to reject than accept PRs that raise this kind of issue.

PR #748, opened 2019-01-03
Addition - Public Bulk Data - Dataverse + Kaggle

lkfitz:
Addition to highlight bulk data availability on Harvard Dataverse and Kaggle.

### DIFF
--- a/capstone/capapi/templates/bulk.html
+++ b/capstone/capapi/templates/bulk.html
@@ -20,6 +20,7 @@
       <p>Click a link to download.</p>
       {% include "includes/download_list.html" with zips=exports.public %}
     </div>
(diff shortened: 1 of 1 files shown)


anastasia:
Tiniest nitpick: would you mind adding target="_blank" to those links so that they open in new tabs instead of taking the user away from the cap site?

bensteinberg:
I think the `p` tag probably needs to be in a `div`, and I'm not sure this is the right page for this content, in any case. Maybe a subsection in https://case.law/bulk/ ?

lkfitz:
Sounds good! I'll whip this up as a subsection on https://case.law/bulk/ and we can revisit.

lkfitz:
Here's my attempt at a subsection! Feedback encouraged https://github.com/harvard-lil/capstone/pull/749

rebeccacremona:
(Closed by request; see #749 )

PR #757, opened 2019-01-18
Reset api key

ChefAndy:
Adds a button to the user details page which brings the user to a little confirmation page ('reset-api-key') with a button to reset the API key. Pressing the button will submit a POST request to that same route, which will reset the key, send the user a confirmation email, and redirect them to their account page.


jcushman (dismissed):
Sorry I'm just getting to this! Can we just tweak this so it doesn't include the old/current API keys in the email confirmation? Some people may prefer not to have sensitive keys in their email box.

ChefAndy:
> Sorry I'm just getting to this! Can we just tweak this so it doesn't include the old/current API keys in the email confirmation? Some people may prefer not to have sensitive keys in their email box.

Good call

jcushman:
Nice!

PR #759, opened 2019-01-29
Package updates

bensteinberg:


bensteinberg:
Updating Django to any 2.1 version requires django-model-utils to cut a release.

PR #764, opened 2019-02-04
Case Searching

ChefAndy:
- Vue-based case search interface at /search/
- Simple starter search docs at /search-docs/
- Updated nav bar


capstone/capbrowse/templates/view_case.html
```diff
@@ -0,0 +1,160 @@
...
+      <div class="row below-the-fold">
+        <div class="main-content">
+            <div v-html="casebody" class="content">
+            </div>
+
+          </div>
+        </div>
+    </div>
+  </div>
+  <script src="https://cdn.jsdelivr.net/npm/vue@2.5.17/dist/vue.js"></script>
```
jcushman: This should be included through our JS pipeline.

capstone/capbrowse/views.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+# Create your views here.
+
+def view_case(request, case_id):
+    return render(request, "view_case.html", {'case_id': case_id})
+
+
+def view_jurisdiction(request, jurisdiction_id):
+    jurisdiction = Jurisdiction.objects.get(pk=jurisdiction_id)
```
jcushman: This line is better off using `get_object_or_404` so it doesn't throw a 500 with an unknown jurisdiction. Same for subsequent functions.

capstone/capbrowse/views.py
```diff
@@ -0,0 +1,57 @@
...
+        'title': court.name_abbreviation
+    })
+
+
+def search(request):
+    return render(request, "search.html")
+
+
+def _get_fields(object):
+    fields = OrderedDict()
```
jcushman: This is a cool function, but dangerous -- if we ever add an internal-use field to one of these models, it will silently leak to the public. Can we switch it to take a list of fields instead of using `object._meta.get_fields`?

capstone/capweb/templates/includes/nav.html
```diff
@@ -35,6 +35,7 @@
           <a class="dropdown-item" href="{% url "tools" %}">Overview</a>
           <a class="dropdown-item" href="{% url "api" %}">API</a>
           <a class="dropdown-item" href="{% url "bulk-docs" %}">Bulk Data</a>
+          <a class="dropdown-item" href="{% url "search" %}">Search</a>
```
jcushman, line 39: Let's not include this endpoint in the nav yet, so we can kick the tires on prod before sharing it.

capstone/capbrowse/templates/search.html
```diff
@@ -0,0 +1,63 @@
...
+    <div class="loading-text">Loading</div>
+  </div>
+  <div id="app" class="content"></div>
+
+
+
+  <script>
+// This is essentially each endpoint with its API interface
+// eslint-disable-next-line
+  app = document.getElementById("app");
```
jcushman: Unused variable?

capstone/capbrowse/views.py
```diff
@@ -0,0 +1,51 @@
...
+    for field in object._meta.get_fields():
+        print(field.name, field.get_internal_type())
```
jcushman: Remove debugging print statements

jcushman (changes requested):
Nice! Some suggestions ...

capstone/capbrowse/templates/view_case.html
jcushman: Never mind -- this file was replaced!

capstone/capbrowse/views.py
ChefAndy: Manually setting fields, so this is gone anyway.

capstone/capweb/templates/includes/nav.html
ChefAndy: Ok

ChefAndy: Commented it out in the template

capstone/capbrowse/views.py
ChefAndy: Got rid of it. Just setting the group of fields to display manually.

capstone/capbrowse/views.py
ChefAndy: Done.

jcushman (approved):

PR #778, opened 2019-02-08
Use TransactionTestCase

bensteinberg:
See https://pytest-django.readthedocs.io/en/latest/database.html#testing-transactions

This allows versioning tests to pass whether with the temporal_tables extension or the plpgsql fallback.


bensteinberg:
Stay tuned. Another test is failing.

PR #781, opened 2019-02-11
First cut at submitting with enter key

bensteinberg:
I *think* this accomplishes the desired effect, viz. allowing "submission" of the search form by hitting the enter key in a text input field.

Closes #777.

PR #782, opened 2019-02-11
Submit search form when pressing enter

### LINKED ISSUE #777: hitting enter in Full-Text Search form field should start search

jcushman:
Second pass at Ben's PR from #781:

- move new-search event from submit button to form
- change remove-field button from type="submit" to type="button" so it doesn't cancel submit event
- also drop quotes from example search string, as in Ben's PR

Closes #777.


capstone/static/js/search/search-form.vue
```diff
@@ -52,14 +52,10 @@
...
               <div class="remfield">
-                <button v-if="fields.length > 1"
-                        class="field-button active"
+                <button type="button"
+                        :class="[fields.length > 1 ? 'active' : 'disabled', 'field-button']"
                         @click="removeField(field['name'])">
                 </button>
-                <button v-if="fields.length <= 1"
-                        class="field-button disabled"
-                        disabled>
```
rebeccacremona, line 61: Absent some behind-the-scenes js, I don't think the class="disabled" and the html attribute "disabled" are the same. Generally you can still tab to and hit enter to submit a button with the css-targeting class "disabled"; generally the html attribute "disabled" prevents both.

jcushman: Whoops, missed the `disabled` attr when DRYing the two versions of the button -- thanks!

capstone/static/js/search/search-form.vue
```diff
@@ -52,14 +52,11 @@
...
-                <button v-if="fields.length > 1"
-                        class="field-button active"
+                <button type="button"
+                        :class="[fields.length <= 1 ? 'disabled' : 'active', 'field-button']"
+                        :disabled="fields.length <= 1"
                         @click="removeField(field['name'])">
                 </button>
-                <button v-if="fields.length <= 1"
-                        class="field-button disabled"
-                        disabled>
```
bensteinberg, line 61: Doesn't this need to be either `disabled` or `disabled="disabled"` (for XHTML), or on the other hand nothing at all?

jcushman: My force-push fixed that -- in Vue, if you return `false` from a bound attribute, the attribute is excluded. `:disabled="fields.length <= 1"` will include `disabled="disabled"` for fields.length <= 1, and otherwise will include nothing.

PR #793, opened 2019-02-14
Added back scrollToResults

ChefAndy:


rebeccacremona:
From the peanut gallery: you might want to consider altering scrollToResults to set keyboard focus instead, at some point. [scrollIntoView)(https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) scrolls the viewport, but leaves keyboard focus unchanged. When you set keyboard focus, the scrolling happens too, automatically.

ChefAndy:
Good call. Thanks @rebeccacremona!

capstone/static/js/search/main.vue
```diff
@@ -138,7 +138,10 @@
           let self = this;
           let url = this.assembleUrl(this.search_url, this.endpoint,
               this.cursors[this.page], this.page_size, this.$refs.searchform.fields);
-          this.getResultsPage(url).then(self.lastFirstCheck);
+          this.getResultsPage(url).then( function () {
+              self.lastFirstCheck();
+              self.scrollToResults();
```
jcushman: Dang, now I see the issue -- the problem with doing this here is that I added a `catch` to the promise chain in `getResultsPage`, to handle connection errors. `then()` methods after a catch will run despite the error, so now this `then()` will run whether or not there's an error, which we don't want -- it scrolls away from the error message.

I think maybe the right thing to do is to do `throw response` at the bottom of the `catch()` in `getResultsPage`, so these handlers outside `getResultsPage` won't run. This would also mean moving the `self.stopLoading();` so it runs whether or not there's an error.

capstone/static/js/search/main.vue
```diff
@@ -237,6 +236,7 @@
 
               self.results[self.page] = results_json.results;
             })
+            .then(self.scrollToResults)
```
ChefAndy, line 239: @jcushman I think this does everything we want it to do— It won't run if you hit the throw, but .then(self.stopLoading) after the .catch does.

jcushman:
Nice!

PR #799, opened 2019-02-19
showing black+whitelisted cases in a similar way

anastasia:
Without the links (which are currently commented out), things look a little drab. But that should be fixed soon.

screen shot 2019-02-19 at 5 57 29 pm

screen shot 2019-02-19 at 5 57 43 pm

### DIFF
--- a/capstone/capapi/renderers.py
+++ b/capstone/capapi/renderers.py
@@ -8,6 +8,8 @@ from rest_framework import renderers
 
 from capweb.helpers import cache_func
 from scripts.generate_case_html import generate_html
+from scripts.process_metadata import parse_decision_date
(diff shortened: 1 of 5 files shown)


anastasia:
Added two more cases:

if parameter is missing — 

screen shot 2019-02-21 at 10 38 09 am

if case is missing —

screen shot 2019-02-21 at 10 38 19 am

There is one more case that isn't covered by this PR: if there is an Authentication Error, we're serving that super simple generated html page. Unsure of how to trigger it in the browser, this seems like a command line only error. Is that right? If it's only in the command line, that generated page seems appropriate. Otherwise I will fix it.

PR #808, opened 2019-02-25
Non-modal loading gif in search; bug fixes

jcushman:
Sorry for the big change dump -- this includes a bunch of little changes I made on Friday/over the weekend as I was working on the loading GIF.

The primary change here is to remove the loading div that blocks the UI, and instead show the loading gif in the results area. To make sure it gets seen, the page smoothly scrolls down when loading starts, and scroll back up in case of an error.

Other smaller changes included:

- Move Django urls to single `urls` variable
- Tweak counts: "No results" or "Results X to Y of N" or "Results X to Y of many" at top of results list
- Tweak case list contents; add human-readable dates
- Use VueScrollTo to smooth-scroll to results page or error messages as appropriate
- Use URL object (and polyfill) to parse URLs; fix cursor encoding bug
- Streamline newSearch/nextPage/prevPage/getResultsPage flow
- Tighten text below search form


rebeccacremona:
Saw this go by in my email, and I have a couple accessibility questions:

  *   Is keyboard focus handled by the “VueScrollTo” moves? If not, you can set keyboard focus manually using the same selectors right afterwards. (I say afterwards, rather than before, to preserve the “smoothness”; setting keyboard focus will itself achieve standard scrolling, if done first.)
  *   Scrolling down, and then back up on error, sounds a little disorienting, and like the kind of unexpected automatic motion that gives people motion sickness. Let’s see if we can put our heads together to design an experience with less motion.

jcushman:
@rebeccacremona -- thanks! Happy to say I had both those issues in the back of my head as things to figure out once I got this far -- but I do have questions about the details. Let's kick it around.

PR #810, opened 2019-02-27
 Use vue-router for tracking state in search URL

jcushman:
Use VueRouter for tracking the search state. All updates (form submit / dropdown change / etc.) now just update the URL hash, and then we update the display based on Vue's parsing of the hash. This gets us free back/forward handling and URL parsing, as well as better handling if the URL is manually edited and re-submitted.

Along the way I switched to a url format like `/search/#/cases?page=num&cursor=foo&fieldname=value`, which VueRouter can easily parse.


rebeccacremona:
A remark from the peanut gallery, about accessibility as usual, and which you may already know. Generally, updating the hash to anything but the id of an element is problem, because the browser is supposed to scroll to/set keyboard focus to the element named by the hash: in page anchors! If an element with the id of whatever follows the `#` doesn't exist, keyboard users generally get ejected to the top of the page, and the viewport scrolls up. I'm assuming none of that bad behavior surfaces here because of all the hard work you did to manually handle keyboard focus and scrolling in your earlier PRs. But thought it was worth mentioning since, when a a developer hasn't done all that work, or doesn't want to, altering the hash is dangerous territory.

jcushman:
@rebeccacremona -- are you sure we're talking about the same thing? It's common for single-page apps to track state via `location.hash = 'foo'` (old way) or `history.pushState({}, '', '#bar')` (newer API), and I can't find anything about either of those changing focus. In fact they don't cause navigation even if they do happen to target existing anchors.

rebeccacremona:
@jcushman Yep! Talking about the same thing. My understanding of the situation could be out of date. I'll find out.

rebeccacremona:
Yep: I am officially way, way out of date. Now, in all three browsers on my mac EVEN FOLLOWING A BROKEN IN-PAGE LINK doesn't cause a problem. Tested with
```
<!DOCTYPE html>
<html>
<head>
    <title>Link with broken hash</title>
</head>
<body>
    <div style="height:300px;"><button type="button">First</button></div>
    <div style="height:300px;"><button type="button">Second</button></div>
    <div style="height:300px;"><button type="button">Third</button></div>
    <a href="#foo">I'm a link to nowhere.</a>
</body>
</html>
```
That definitely used to reset focus and scroll: I can recall personally encountering it as recently as.... 2012? 2013?

I think that old browser behavior, plus use of `window.location = "#foo"` (and possibly `window.hash`....) was the problem with early single-page apps....

But yes, this appears to no longer be relevant!!! Which is a true delight.

PR #815, opened 2019-03-05
Load ngram text files into postgres

jcushman:
Here's a pass at loading the ngram .xz files into postgres, with a threshold of 10 instances or more.


jcushman:
I'm going to go ahead and pull this so I can try it out on beta.

PR #817, opened 2019-03-06
add printedpagenumber (printed page number) attribute to html case elements 

ChefAndy:


jcushman:
Documenting chat with @ChefAndy in lilspace: I think we're currently leaning toward `<a class="page-number" id="p798" href="#p798" data-label="798" data-citation-index="1">798*</a>`.

abziegler:
Confirming chat with @chefandy on slack - we're moving the asterisk to before the page number, in order to follow convention - so like this: `<a class="page-number" id="p798" href="#p798" data-label="798" data-citation-index="1">*798</a>`

PR #818, opened 2019-03-06
Fix file mode; make ngram load resumable

jcushman:


jcushman:
Merging to run on beta.

PR #821, opened 2019-03-07
Reorder head matter in API views

jcushman:
Here's a first pass at addressing the issue we talked about in Slack today. As @ChefAndy suggested, we can use the `<structMap TYPE="xref">` part of casemets to re-order headmatter blocks by the order they appear in the ALTO, which should map to reading order.

To keep things simple, I just changed extract_casebody() to do the reordering, and made sure we were calling that for all of our API views. So this only changes the appearance of cases when we display them, rather than permanently storing the reordering anywhere. If we like, we can do something more efficient/permanent later.

On the tiny number of cases I've looked at, this seems to be a definite improvement.

**Example one:**

Old way:

image

New way:

image

In this example the old version is totally misleading -- the text paragraphs at the end are the arguments of the parties, and only make sense with the attorney paragraphs in the right locations.

**Example two:**

Old way:

image

New way:

image
 
Here the old way had a weird thing where it looked like the case name was in brackets for some reason, when actually the case just had "[No. 33971-8-III. Division Three. July 11, 2017.]" printed at the top.

**To do later:**

* Spot check more examples on beta.
* Tweak styles based on text appearing in a less predictable order.
* Remove the extra linebreaks that are slipping in because single paragraphs are split into two. I think maybe we can just turn `id="A..."` head matter elements into spans in the HTML generator.

### DIFF
--- a/capstone/capapi/serializers.py
+++ b/capstone/capapi/serializers.py
@@ -167,14 +167,15 @@ class CaseSerializerWithCasebody(CaseAllowanceMixin, CaseSerializer):
 
         if casebody['status'] == 'ok':
             # if status is 'ok', we've passed the perms check and have to load orig_xml into casebody['data']
-            orig_xml = case.case_xml.orig_xml
 
             # non-JSON, single-case delivery formats will be handled by custom renderers
             if type(request.accepted_renderer) == HTMLRenderer:
-                data = orig_xml
+                data = case.case_xml.extract_casebody()
                 casebody['title'] = case.full_cite()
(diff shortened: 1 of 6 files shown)


jcushman:
Oh, we should check a case that has footnotes in the head matter! Not sure if those should be reordered.

ChefAndy:
Nice turnaround time on this! I’ll check it out when I get in tomorrow.

ChefAndy:
Whatever we do with spans vs paragraphs, we want to be sure that we communicate with H2O about it. Adding an easily strippable style that displays it as inline-block might be less invasive.

PR #834, opened 2019-03-15
Styled XML— Model change and test, fab task

ChefAndy:
Changed the CaseXML model to include a field for styled XML, and a method to update it.
Also added a test for that stuff.
Added a few arguments to the generate styled XML function to specify skipping dups
without raising an exception, and also appended a comment to the generated XML which
says if the text didn't totally match up with the ALTO.
Added a fab task which loops through each volume, synchronously, and asyncrhonously
updates the styled text for all of the cases in that volume.


capstone/capdb/models.py
```diff
@@ -856,6 +857,8 @@ class CaseXML(BaseXMLModel):
                                on_delete=models.DO_NOTHING)
     s3_key = models.CharField(max_length=1024, blank=True, help_text="s3 path")
 
+    styled_case_body = XMLField(blank=True, null=True)
```
jcushman: Let's do this as a separate table, like `CaseText` (which is another cache of case output). There's a bunch of stuff that might assume that fetches from `CaseXML` are fast if they don't include `orig_xml`.

capstone/fabfile.py
```diff
@@ -928,6 +928,28 @@ def sample_captar_images(output_folder='samples'):
                 )
 
 
+@task
+def update_all_styled_xml_casebody(skip_existing=True):
+
+    @shared_task
```
jcushman: I don't think this task will get discovered by celery when it's defined here.

capstone/fabfile.py
```diff
@@ -928,6 +928,28 @@ def sample_captar_images(output_folder='samples'):
...
+    def update_styled_xml(case_xml):
+        try:
+            case_xml.update_styled_casebody(save_self=True, skip_duplicative=True, strict=False, body_only=True)
+            return {case_xml.metadata.case_id.split('_')[-1]: "ok"}
+        except Exception as e:
+            return {case_xml.metadata.case_id.split('_')[-1]: e}
+
+    for volume in VolumeXML.objects.select_related('metadata').all():
+        job_list = group([ update_styled_xml.s(case_xml) for case_xml
```
jcushman: This doesn't really work with celery. Whatever input you queue for a job has to be serialized, stored in redis, and then pulled off and deserialized by the workers. So the job parameters should be json-serializable and small, like IDs, rather than Django ORM objects. It's also a little hairy to queue millions of jobs and wait for the result -- should work in theory, but may not in practice.

Check out `create_case_metadata_from_all_vols` and `create_case_metadata_from_vol` in tasks.py for a good example of how to handle this. Going per-volume breaks the job into 40,000, which is more manageable, and allows us to fetch data efficiently from the ORM for each volume. If you need to gather errors, it's probably a good idea to write them to the log from celery (I think just print), or write them to a table -- it's not reliable to assume the original task will still be running when all the subtasks are complete.

To get a feel for this stuff it might be worth getting celery running locally -- I think Ben said the Docker setup is currently working for him.

jcushman (changes requested):
A couple of suggestions on the storage and processing side ...

capstone/capdb/models.py
ChefAndy: ok, makes sense

capstone/fabfile.py
ChefAndy: ah, I see. I'll see if I can fix my horked docker setup and work in a more representative environment.

PR #835, opened 2019-03-18
fonts

anastasia:
- moved all font face definitions to fonts.scss
- added a bunch of new fonts (too many? I don't *think* it should be a problem but I could be wrong!)
- small text font should be source sans
- all serif texts except for the big quote on index page should be source serif


jcushman:
That does look like a loooot to load, but let's try it and see what it adds up to on first load in a browser ...

PR #840, opened 2019-03-22
Search design updates

jcushman:
This includes some work from me and from @anastasia.

Compared to the design mocks the biggest intentional change is to move the full text search help text to right under the FTS field instead of on the right -- I found that it was pushing the results down a lot otherwise, and it seemed nice to put it closer to the field it was explaining.

On the code side, I started using the bootstrap breakpoint classes instead of custom classes for setting breakpoints. So instead of this:

```
@include media-breakpoint-up(lg) {	
  .results-list-container {	
    @include make-col(8)	
  }
}
@include media-breakpoint-up(sm) {	
  .results-list-container {	
    @include make-col(10)	
  }
}
```

It's this:

`<div ... class="results-list-container col-lg-8 col-sm-10">`

I'm finding the second style much easier for quickly tweaking breakpoints in a grid.

Stuff that still isn't done:

- dropdown caret for endpoint dropdown (I would love to just move us to boostrap dropdowns instead of custom, but haven't)
- plus sign for add field button
- any styling for other endpoints (I killed some of what was there along the way, so they now just list facts in order)

image


anastasia:
I'm going to add a couple more tweaks

anastasia:
Actually, I take that back. Merging now and will add more tweaks in another PR

PR #844, opened 2019-03-25
Revert "Styled XML— Model change and test, fab task"

bensteinberg:
Reverts harvard-lil/capstone#834 temporarily, for deployment purposes.


bensteinberg:
Never mind.

PR #848, opened 2019-03-25
Ordered Search Choices

ChefAndy:
I changed the snippets to be arrays of arrays rather than dicts. It's not quite as elegant but it preserves the order both in python and in the browser in JS.


ChefAndy:
I think this is a reporter short name sorting difference between travis and macos.

ChefAndy:
Don't merge this— I have no idea why the test is failing on travis but working locally.

ChefAndy:
My local PG implementation sorts `Reporter: Ill. (Breese): Illinois Reports 1819-1831` in front of `Reporter: Ill. App.: Illinois Appellate Court Reports 1877-1887` while the docker instance consistently sorts them the other way around.

PR #867, opened 2019-03-28
Fix up CAP search gallery

bensteinberg:


bensteinberg:
Hover behavior on one link is weird, investigating... never mind, looks fine.

ChefAndy:
Travis is still being lame, but this is HTML and an image. I think we're good.

PR #871, opened 2019-03-28
Extract data from CAPTAR and store in new tables as JSON tokenstreams

jcushman:
Sorry for the megapull -- more details via Google Docs/Slack

Significant user-visible changes if this is pulled:
- HTML pages put the head matter in `<section class="head-matter">` block
- /cases/ID/?format=xml&full_case=true returns only <casebody>, not entire CaseMETS


capstone/capdb/models.py
```diff
@@ -842,12 +856,169 @@ def create_or_update_case_text(self, new_text=None):
...
+        ## create json
+
+        casebody_pq = PyQuery(html)
+
+        # add footnote labels to footnotes
+        for footnote in casebody_pq('.footnote').items():
+            label = footnote.attr['data-label']
+            if label:
+                footnote_paragraph = footnote.children()[0]
+                footnote_paragraph.text = "%s %s" % (label, footnote_paragraph.text)
```
bensteinberg, line 902: Do we ever do anything with `footnote_paragraph`? I can't see where.

capstone/scripts/refactor_xml.py
```diff
@@ -0,0 +1,936 @@
...
+            page_obj = PageStructure(blocks=page['blocks'])
+            page_obj.encrypt(key)
+            page['encrypted_strings'] = page_obj.encrypted_strings
+
+        # create an unencrypted copy for checking reversability
+        print("Decrypting pages")
+        decrypted_pages = deepcopy(pages)
+        for page in decrypted_pages:
+            page_obj = PageStructure(blocks=page['blocks'], encrypted_strings=page['encrypted_strings'])
+            page_obj.decrypt(key)
```
bensteinberg, line 766: Do we just decrypt to make sure it runs, or do we check the results in `page_obj` as well? I think I'm probably missing something.

capstone/capdb/models.py
jcushman: This modifies the value of casebody_pq, which then shows up in opinion.text() down below.

capstone/scripts/refactor_xml.py
jcushman: This decrypts the pages in decrypted_pages. Down below in `for page in decrypted_pages` we make sure we can reconstruct each ALTO <Page> element, which relies on the decryption succeeding.

PR #875, opened 2019-04-03
Remove django-partial-index

bensteinberg:
This should be a no-op -- we're creating the same indexes manually that were previously created with partial_index. If we need to make any similar indexes in the future, we can do it by hand, or experiment with `condition` when we get to Django 2.2.


bensteinberg:
Thanks for the help with this, @jcushman.

jcushman:
Nice!

PR #887, opened 2019-04-08
Add cite subdomain with jurisdiction/series/volume/case viewer, using cookie wall

jcushman:
OK! Here's a case browser with readable links and a cookie wall. This is not ready for primetime (see TODO), so I have it behind a feature flag, but it would be cool to try it out on beta with real data.

cite.case.law:

image

cite.case.law/ill-app/:

image

cite.case.law/ill-app/23/:

image

cite.case.law/ill-app/23/19/:

image

**Feature flag:**

`CITE_SUBDOMAIN_FEATURE=True` (which is true in dev, false in prod)

**Routes:**

`/` -- list of all jurisdictions and reporters
`/<series_slug>/` -- list of all volumes for given series
`/<series_slug>/<volume_number>/` -- list of all volumes for series
`/<series_slug>/<volume_number>/<page_number>/` -- contents of requested case, or list of multiple cases with that cite, or link to Google Scholar if the cite isn't found[](url)
`/<series_slug>/<volume_number>/<page_number>/<case_id>/` -- contents of requested case, if necessary to disambiguate

**Cookie wall:**

The goal of the cookie wall is that you should have to deliberately circumvent the cookie wall (and thus violate the terms of use) if you want to exceed the quota.

We use this logic:

- If case is whitelisted, display
- If user is logged in, use their normal quota
- If user is not logged in, check if they have a `not_a_bot=1` cookie and a django session cookie
  - If yes, track their case limit in their django session
  - If no:
    - First try serving a javascript page that submits a POST form to set the not_a_bot=1 cookie
    - If javascript fails, serve a regular HTML page to submit the POST form to set the not_a_bot=1 cookie

The upshot is that our page checks that users support cookies, and either: (a) are running javascript, or (b) are manually clicking a button to agree, or (c) are bot authors who are manually setting a 'not_a_bot' cookie. So if you're a bot author, you have to either set a cookie manually, or use a headless browser and deliberately throw away cookies every 500 requests, which are both overt circumventions.

**TODO:**

- Confirm routes and cookie wall implementation with team.
- Tests.
- Design.
- Add banner on first view that says "welcome, you get 500 cases per day, please follow the terms of service."
- Add some way to look up cites from "<volume_number> <reporter short name> <page_number>" form -- I think that's an additional route like 'cite.case.law/<raw_cite>', plus a form on 'cite.case.law' landing page.
- Confirm TOS covers everything we need it to.
- Set up Cloudflare to show an IP-based rate limit CAPTCHA.


jcushman:
This also clearly exposes the issues with our Reporter data model -- the way we're not differentiating nominative reporters, for example. That'll be something to solve at some point as well.

jcushman:
The redirect also kills pincite links like #p123. Have to think about whether we can avoid that.

PR #902, opened 2019-04-10
Add elasticsearch in dev

bensteinberg:


bensteinberg:
I set this to 6.7.1, not 7.0.0, since that's what we have on beta, but we can change that.

PR #916, opened 2019-04-19
Elasticsearch proof of concept

ChefAndy:
This definitely isn't ready for prime time. The basics are in place, but features are missing, and there are no tests.
Everything can pretty easily be changed and I don't think there's a whole lot left to do to polish it up.

In general, where something was labelled 'Case' in the API code, I labelled it 'CaseDocument' to distinguish them.
If we end up using both in the API in the future, something that more obviously distinguishes them would be better.

* `capapi/documents.py`
    * New file. Should contain our search index definitions. This is the code that defines how our models are translated to Elasticsearch documents and is run when you run the reindexing task.

* `capapi/serializers.py`
    * New serializers
        * `CaseDocumentSerializer` and for non-casebody results
        * `CaseDocumentSerializerWithCasebody` for with-casebody cases.

* `capapi/permissions.py`
    * Standalone function, `check_update_case_permissions`, to check and update case allowances.
    * Changed the list of statuses to a dictionary for readability.

* `capapi/api_views.py`
    * CaseDocumentViewSet to view cases from the `elastic_cases` endpoint. This is where you configure things like pagination, search fields, etc.

* `capapi/api_urls.py`
    * Created the `elastic_cases` endpoint behind a feature flag

* `config/settings/settings_base.py`
    * Created the setting that allows the django DSL to talk to the elasticsearch container.
    * Created the `ELASTICSEARCH_API_ENABLE` setting that toggles the elastic_cases endpoint

* `capstone/requirements.in`
    * Added `django_elasticsearch_dsl`, which includes the indexing/comm stuff
    * Added `django_elasticsearch_dsl_drf`, which includes the serializer/view stuff

* `capstone/requirements.txt`
    * Changed the elasticsearch version back to 6.7.1 because none of the current software supports 7

* `capstone/docker-compose.yml`
    * Doubled ram allocated to elasticsearch container because it was dying with OOM errors.

* `capstone/fabfile.py`
    * Made reindexing fab task.

The general workflow for adding elasticsearch endpoints is pretty similar to adding regular DRF endpoints— 
you define your indexes in `documents.py`, run the indexing task, add a serializer, then a view and a URL. Make sure you
disable the automatic indexing of model changes in `documents.py` because it not only messes up the models, running tests 
will create new entries in the ES index. I'm sure there's some way to get around this if we decide to do it in the future.


jcushman:
Awesome! This looks good overall -- I'm inclined to pull it in and try it out on beta. Just a couple of questions first --

Will this break anything for machines that don't have Elasticsearch running, assuming `ELASTICSEARCH_API_ENABLE=False`? It's probably fine, but for example, does Django try to connect to ES on startup, or just when the route is called?

Do we have to store the XML and HTML in ES? That seems like a ton of redundant storage. I'm not sure if we should try it like this first, or comment that out and try just text first to check performance and then find a way to mix HTML/XML back in through separate DB calls.

ChefAndy:
@jcushman — yeah, I'm totally on the fence about what's included in the elasticsearch documents.  My initial thinking was that we could totally avoid hitting the database for anything that touches the case endpoint, which would mean we could kill the HTML cache and text case text models. Also, if someone gets a page of results for which they want to get HTML or XML bodies, it might take a bit of work to set up the code to get the XML for each page of results in one query, though I'm sure it's doable. Either way, I figured I'd throw what I had up and we could all discuss the architectural stuff together. At this point, it's really easy to change it up.

It works fine in an environment without elasticsearch up— in an environment with elasticsearch, it does require running the index rebuilding fab task before it works. I should probably note that somewhere.

jcushman:
Cool -- let's do it. I was thinking we needed the non-ES HTML and XML for other usecases like the single case page and bulk export, but those actually pull from the API, so maybe not.

bensteinberg:
@ChefAndy, thanks for fixing what I busted. I just bumped version numbers in docker-compose.yml and am going to merge.

PR #920, opened 2019-04-22
Add a mechanism for notifying an external service on task completion

bensteinberg:
I'm not sure this is a great idea, or a great implementation, but it's working for me in other contexts.


bensteinberg:
Whoops, I should add `requests` to requirements.in.

PR #931, opened 2019-04-26
Add pip-compile fab task; include hashes in requirements.txt

jcushman:
For security, this includes the known hashes for all of our requirements in requirements.txt. I added a fab task to wrap `pip-compile` to work around an incompatibility with `pip`.


jcushman:
Forgot to mention this did bump the version of pip-tools from 3.3.2 to 3.6.1. Updated docker image version.

PR #933, opened 2019-04-30
Markdown templates

jcushman:
Just a little PR for fun. This is an attempt to make it simpler for us to keep docs up to date by letting us write them as markdown with automatically generated tables of contents. I converted the terms of use to markdown as an example.

The one limitation that I know of is that markdown docs don't have `<div class="page-section">` in them, which reduces the space before `<h2>` tags. We could possibly move that padding elsewhere -- not sure of the ideal approach.


jcushman:
One neat thing here is we can use "meta" lines at the top of the markdown document to fill in various spots in our base templates, so we don't need any html templates at all for a straightforward doc like the terms of use:

urls.py:

`    path('terms', MarkdownView.as_view(template_name='terms-of-use.md'), name='terms'),`

terms-of-use.md:

```
title: Terms of Use
meta_description: Caselaw Access Project Terms of Use
explainer: Effective: <span class="highlighted">April 23, 2019</span>

 Caselaw Access Project is ...
```

bensteinberg:
This is cool, thanks.

PR #939, opened 2019-05-02
Style fixes

jcushman:
Some more style fixes:

- Fix problems with search page caused by sidebar and padding tweaks
- Add allowedHosts to node dev server so it works on cite.case.test:8000
- Use JS to make sure URL targets don't get covered up by navbar. I couldn't get the pure CSS solutions at https://stackoverflow.com/questions/4086107/fixed-page-header-overlaps-in-page-anchors/ to actually work.


jcushman:
A nice plus of the JS solution for navbar cover-up is it also works on the footnote references in our single-case page.

PR #952, opened 2019-05-09
Update pip-tools and simplify pip_compile fab task

jcushman:
This updates pip-tools to take advantage of the new URL code. pip-sync will hopefully work again for deploys if you want to use it.


bensteinberg:
Cool! I'll put pip-sync back and try it on the next deploy. Will you bump the image number in docker-compose.yml?

PR #967, opened 2019-05-15
Linked data for case.html

anastasia:
- "Article" seems to be the closest match for our linked data object for case law.
https://developers.google.com/search/docs/data-types/article
- `isAccessibleForFree` is set to false: 
https://developers.google.com/search/docs/data-types/paywalled-content
- Courts are our "Author" (with type Organization)
- Date modified is set to our public release date for CAP (October 29, 2018), although maybe it should be more specific?
- keywords are all the citations and the name abbreviation (could be more entries?): https://schema.org/keywords
- Image seems to want to be a specific-to-the-article. Right now I'm only including our social media image. Maybe there's some cool way to include an image of a case? If not, and this seems fine, specifications recommend this image in various sizes so I will add that

To test
View any case, open dev tools, search for `application/ld+json` in the DOM, copy paste the object here:
https://search.google.com/structured-data/testing-tool?utm_campaign=devsite&utm_medium=jsonld&utm_source=article

Screen Shot 2019-05-15 at 11 33 54 AM

### DIFF
--- a/capstone/capapi/renderers.py
+++ b/capstone/capapi/renderers.py
@@ -59,6 +59,7 @@ class HTMLRenderer(renderers.StaticHTMLRenderer):
             'citation': official_citation,
             'page_description': data['name'],
             'page_title': data['name_abbreviation'],
+            'frontend_url': data['frontend_url'],
             'metadata': {
                 "name": data["name"],
                 "name_abbreviation": data["name_abbreviation"],
--- a/capstone/capapi/templates/case.html
+++ b/capstone/capapi/templates/case.html
@@ -1,6 +1,8 @@
 {% extends "layouts/centered.html" %}
 {% load pipeline %}
 {% load static %}
+{% load capweb_static %}
+
 {% block base_css %}{% stylesheet 'case' %}{% endblock %}
 {% block title %}{{ citation_full }}{% endblock %}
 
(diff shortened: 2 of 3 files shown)


jcushman:
Cool!

- Once we have page images available, we could use those as images. For now maybe it's better not to include anything as the image, because it's not article-specific.

- Can we add special handling for whitelisted cases? Maybe what's simplest is to use .casebody as the cssSelector instead of adding .cookie-wall, and then you can just omit the whole "hasPart" section for whitelisted cases.

The other half of this is to add googlebot detection in the view so they don't get blocked by the JS check -- are you thinking it's easier to handle that as a separate PR?

anastasia:
- the image field is unfortunately required
- whitelisted cases — that makes sense! 
- different PRs make sense to me! 

Maybe worthwhile to try out this tool as well (when we set this up, if not before): 
https://search.google.com/search-console/welcome

It requires us to upload some text to the DNS configuration.

anastasia:
ok doing this 

Screen Shot 2019-05-15 at 12 18 00 PM

in the middle of a json ld object feels incredibly wrong to me but it works, and the alternative is doubling up on the data (if and else surrounding the object) and potentially falling out of sync with the structured datasets

what say you, oh great Github Reviewers?

jcushman:
Looks right to me! I don't think templating JSON is really any weirder than templating HTML.

PR #971, opened 2019-05-16
npm updates

jcushman:
Some misc javascript build tweaks:

- Include filename hashes in the JS we serve, by deleting 'filenameHashing: false'. I thought hashes  would be added by whitenoise, but they're not.
- Update some packages in an attempt to reduce warnings during build.
- Use `npm ci` in a few places instead of `npm install` -- particularly on travis.


rebeccacremona:
I was just reading about `npm ci` to see if it would be useful in Dockerland.
> If a node_modules is already present, it will be automatically removed before npm ci begins its install.

Interesting that they think that is a good idea for CI environments... I would have thought that would slow down build a LOT

PR #972, opened 2019-05-17
Cite

anastasia:
- added checking if google bot functionality
- if it is, sending request through and showing full case text, along with schema specifying this content isn't free
- added checks for google bot in permissions, and cite/views

### DIFF
--- a/capstone/capapi/permissions.py
+++ b/capstone/capapi/permissions.py
@@ -1,3 +1,4 @@
+from capweb import helpers
 from rest_framework import permissions
 
 staff_level_permissions = [
@@ -6,6 +7,7 @@ staff_level_permissions = [
     'capapi.change_capuser',
     'capapi.delete_capuser',
 ]
+
 
 class IsSafeMethodsUser(permissions.BasePermission):
     def has_permission(self, request, view):
@@ -34,11 +36,13 @@ def get_single_casebody_permissions(request, case):
     updating case download permissions for user if case is blacklisted
     """
     casebody = {"status": None, "data": None}
-
     if not case.jurisdiction_id:
         casebody["status"] = casebody_permissions["unk"]
 
     elif case.jurisdiction_whitelisted:
+        casebody["status"] = casebody_permissions["ok"]
+
+    elif helpers.is_google_bot(request):
         casebody["status"] = casebody_permissions["ok"]
 
     elif request.user.is_anonymous:
--- a/capstone/capweb/helpers.py
+++ b/capstone/capweb/helpers.py
@@ -1,5 +1,7 @@
 import json
 import re
+import socket
+from bs4 import BeautifulSoup
 from collections import namedtuple
 from contextlib import contextmanager
 from functools import wraps
@@ -7,6 +9,8 @@ import markdown
 from markdown.extensions.toc import TocExtension
 import requests
 import django_hosts
+from django_user_agents.utils import get_user_agent, parse
+from ipware import get_client_ip
 
 from django.conf import settings
 from django.contrib.auth.decorators import user_passes_test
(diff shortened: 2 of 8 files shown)


capstone/capweb/helpers.py
```diff
@@ -172,9 +187,37 @@ def render_markdown(markdown_doc):
...
+    meta = {k: ' '.join(v) for k, v in md.Meta.items()}
     return html, toc, meta
+
+
+def is_google_bot(request):
+    """
+    from https://blog.majsky.cz/detecting-google-bot-python-and-django/
+    """
+    user_agent = get_user_agent(request)
+    if not user_agent.is_bot:
```
jcushman, line 203: This seems simpler to do without the 3rd party library as `if "Googlebot" in request.META.get('HTTP_USER_AGENT', ''):`. That will also avoid running this check for non-Google bots.

capstone/capweb/helpers.py
```diff
@@ -172,9 +187,37 @@ def render_markdown(markdown_doc):
...
+        return False
+    ip, _ = get_client_ip(request)
```
jcushman, line 205: We want to specifically get the IP address from `request.user.ip_address`. This makes sure that we're getting the correct IP address, via proxies, and also that the response will not be cached. Unfortunately setting up the request.user object will complicate one of your tests a bit.

(IP address checking is tricky in general, because requests arrive to Django via multiple proxy servers, so there's no one standard place that the true user IP lives, and if you check the wrong place you can actually let an attacker control the answer. It's best to check in with Ben about how the system is set up. On CAP it ends up in the HTTP_X_FORWARDED_FOR header at the moment.)

(Caching is also tricky -- our middleware marks pages as not-cachable-by-cloudflare if user attributes have been accessed, so getting ip_address from there should make sure the caching behavior is correct. I *think* the current code won't cache google requests anyway, just because request.session is checked first, but good to not make that depend on something unrelated to the googlebot check itself.)

capstone/capweb/helpers.py
```diff
@@ -172,9 +187,37 @@ def render_markdown(markdown_doc):
...
+    except (socket.herror, socket.error):
+        return False
+    domain_name = ".".join(host.split('.')[1:])
+    if domain_name not in ['googlebot.com', 'google.com']:
+        return False
+    host_ip = socket.gethostbyname(host)
+    return host_ip == ip
+
+
+def get_schema(response):
```
jcushman, line 218: This seems pretty specific to your test logic -- maybe it wants to live in test_views.py?

capstone/capapi/permissions.py
```diff
@@ -34,13 +36,15 @@ def get_single_casebody_permissions(request, case):
...
     """
     casebody = {"status": None, "data": None}
-
     if not case.jurisdiction_id:
         casebody["status"] = casebody_permissions["unk"]
 
     elif case.jurisdiction_whitelisted:
         casebody["status"] = casebody_permissions["ok"]
 
+    elif helpers.is_google_bot(request):
```
jcushman, line 45: We don't want to add this test here -- that will also open up API queries. Instead we want to use `serializers.NoLoginCaseSerializer` down in cite/views.py.

capstone/cite/tests/test_views.py
```diff
@@ -149,3 +157,75 @@ def test_single_case(client, auth_client, django_assert_num_queries, case):
...
+    # our bot has seen too many cases!
+    session = client.session
+    session['case_allowance_remaining'] = 0
+    session.save()
+    assert session['case_allowance_remaining'] == 0
+
+    case.jurisdiction.whitelisted = False
+    case.jurisdiction.save()
+
+    response = client.get(url, follow=True)
```
jcushman, line 223: Let's add an `assert not is_cached(response)` in here to make sure the response isn't getting cached.

jcushman (changes requested):
Good progress! Some tweaks to make ...

capstone/capweb/helpers.py
anastasia: Ah, I see! I also see the wrap_user function now. Fixing

PR #997, opened 2019-05-24
Ngram API updates

jcushman:
- Use LMDB to serve the ngram API -- hopefully much faster
- New ngram API format -- observations look like this:

```
{
                    "year": "1819",
                    "count": [
                        25,
                        319
                    ],
                    "doc_count": [
                        1,
                        1
                    ]
                }
```

where the word constituted 25/319 total words, and appeared in 1/1 total documents for that jurisdiction-year.

- When searching for a wildcard, like "the best *", API will return top ten results in decreasing order.
- Ngram viewer:
  - Use new API format
  - Show separate line for each result in a wildcard search
  - Show separate line for each result in a multi-jurisdiction search


jcushman:
Whoops, have to fixup something, hang on a sec.

PR #1019, opened 2019-06-03
Ngram viewer updates

jcushman:
This PR has a bunch of random changes to the ngram frontend viewer that I was working on the end of last week. Some are definitely a good idea, and there's one or two that Anastasia may want to nix. :)

Updates:

- Drop redundant CSS file.
- Make chart bigger on wider screens (10 columns instead of 7).
- Graph rendering:
  - Avoid a blank graph with a 0-1 y axis on first load.
  - Use colorblind-friendly and high-contrast-friendly color palette.
  - Only show points on the graph if there are fewer than 50 years being displayed, so they don't all blend together.
  - If points are displayed, use different point styles for each line to help with distinguishing them.
  - Don't use a fill color under the lines, so multiple lines can be seen at once.
- Date range display:
  - Increase default end year to 2018.
  - Zoom in to actual date range returned (so we don't start at 1800 for "ipod").
  - Insert zeroes for years that don't have any data.
  - Store full data returned, so we can update year display without refetching.
- Routing/fetch logic:
  - Add loading indicator to Graph button.
  - Add error messages for searches that don't have any results.
  - Use the route update event to trigger graph redrawing, to simplify setting and retrieving app state.
  - Send queries for all terms in parallel (which is faster, and makes it easier to set correct date range).
  - Don't change graph view unless data is successfully returned (which avoids a weird rendering bug when running a failed search followed by a successful search).
- Add links for example searches under search box.
- Default to graphing percentage of cases that contained the gram, instead of absolute instance count.
- Add an Options panel to let you control:
  - Range of years displayed (which is now instant instead of triggering a new web request)
  - Case count or instance count
  - Percent or absolute count
  - Ability to average out years with nearby years
- Update tooltips to reflect currently selected options, like "2000-2005: apples: 2% of cases"
- Instead of jurisdictions dropdown, search for jurisdictions by prepending the jurisdiction code to the ngram. This lets you compare different terms from different jurisdictions, like "mass: red sox, ny: yankees".
- Use "*: apples" to see "apples" in all jurisdictions separately.
- Add a help panel to see the list of all the jurisdiction codes.

Screenshot:

image

Options panel:

image

Help panel:

image

Ticks in zoomed-in view:

image


anastasia:
This is beautiful!

PR #1036, opened 2019-06-11
Update Django to 2.2.2

bensteinberg:


bensteinberg:
I had to run `docker-compose run web yarn build` to get tests to pass locally; it's not clear to me why this is the case.

bensteinberg:
Trying without the spurious (?) JS changes....

PR #1045, opened 2019-06-14
Styling Updates and Markdown

ChefAndy:
Ok, so I broke this off from the other part of the initiative to standardize the layout to use all native bootstrap classes— that needs a bit more consideration and planning that I expected when I started out.

In this PR, I implemented the new design for all of the rest of the pages based on the design specs, converted most of the pages in which people would need to edit the textual content to markdown, with the notable exception of the API docs which has some design elements that will need to be rethought for markdown, and the gallery which will take a different direction entirely. There's a small extension for Python Markdown in there which, in conjunction with the attr_list extension, allows you to add a class to a list which it doesn't support.


ChefAndy:
have some changes to make

ChefAndy:
Made changes and converted API docs to MD.

jcushman:
Thanks -- big chunk of work converting all that stuff to markdown!

PR #1049, opened 2019-06-18
RocksDB ngram loader

jcushman:
Experimental ngram loader based on RocksDB:

- tweak the tokenizer to strip common punctuation from the front and back of each token
- load ngrams from CaseBodyCache.text instead of CaseText.text
- add feature flag `settings.ROCKSDB_FEATURE`, with which:
  - `fab ngram_jurisdictions` starts celery tasks that stream results directly into a RocksDB key-value store instead of starting a multi-step file-based process
  - ngrams API returns results directly from the RocksDB store

This can run once our currently-running(?) `fab refresh...` command is done.


jcushman:
Aw look at my nice fast-fail for flake8.

PR #1053, opened 2019-06-20
Fix multi-cite case linked by ID

jcushman:


jcushman:
This will hopefully fix https://cite.case.law/f-cas/15/1030/6653167/

PR #1055, opened 2019-06-20
Change log & Ngrams API endpoint docs

ChefAndy:


ChefAndy:
(just stub content for the change log)

PR #1068, opened 2019-06-28
Update models to split out nominative reporters

jcushman:
This is a first step to clean up our nominative-reporter data model -- just adding some additional fields. Next step will be to try a dry-run data update script on read-only beta database.


jcushman:
(we're holding this until after the next deploy)

PR #1072, opened 2019-07-03
update - changing default command to bash to allow for python debugging

anastasia:
this is a suggestion PR!

Wondering what people think about making the docker flow a bit more extensive for the trade-off of being able to debug with pdb/ipdb (this is like the perma.cc flow -- thanks @rebeccacremona for the help)
The flow would be:
```
docker-compose up -d # run docker in background
dfab run:0.0.0.0:8000 # run django in container
dyarn serve # compile front-end assets
```

If this looks good to people, I would need to edit the README to reflect a different workflow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ staticfiles/
 .sass-cache/
 .idea/
 settings.py
+docker-compose.override.yml
 __pycache__
 .vagrant/
 .virtualenvs/
--- a/README.md
+++ b/README.md
@@ -142,20 +142,38 @@ Capstone should now be running at 127.0.0.1:8000.
 
 We have initial support for local development via `docker compose`. Docker setup looks like this:
 
-    $ docker-compose up &
+    $ docker-compose up -d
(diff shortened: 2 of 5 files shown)


jcushman:
This seems fine to me -- let's do it (when you've made any other changes you want to make).

A few suggestions for either the README or fabfile to support this workflow:

- rather than have shortcuts like `dfab` and `dyarn` you could just have a shortcut `d` that calls `docker-compose exec web`, so you can run any command like `d fab`, `d yarn`, `d ./manage.py` etc.
- the `fab run` command could default to 0.0.0.0:8000 when DOCKERIZED=True, so it's just `d fab run`
- it can be handy to start and stop yarn serve and fab run at the same time, which I think would be something like `d bash -c "yarn serve & fab run"`, though I haven't tried it in Docker

anastasia:
btw `d bash -c "yarn serve & fab run"` works in docker. Going to add an alias for speediness in my bash_profile

PR #1074, opened 2019-07-03
Elasticsearch updates

jcushman:
Playing around with our experimental Elasticsearch endpoint:

- Add some README notes on debugging.
- Update the document schema so `casebody_data` is not a nested document. This will let us query `text` and other fields with `simple_query_string`. Also tweak some other fields. This will require reindexing.
- Update indexing code to draw contents from `body_cache`, and to run in parallel celery workers.
- Add cursor-based paginator.
- Fix full_case=true response
- Avoid fetching unused data when preparing response
- Tweak settings so `fab rebuild_search_index` works outside Docker


jcushman:
I'm going to cheat and pull this so it can index on beta over the weekend. :)

PR #1088, opened 2019-07-11
Cite 404

anastasia:
Screen Shot 2019-07-11 at 2 00 34 PM


capstone/cite/views.py
```diff
@@ -198,10 +198,18 @@ def citation(request, series_slug, volume_number, page_number, case_id=None):
                 .prefetch_related('citations'))
         else:
             cases = []
+        reporters = list(Reporter.objects.filter(short_name_slug=slugify(series_slug)))
```
jcushman: This can be a little cleaner with `first()`, which returns `None` if not found:

```
reporter = Reporter.objects.filter(short_name_slug=slugify(series_slug)).first()
series = reporter.short_name if reporter else series_slug
```

jcushman:
Cool! Just added one suggestion re: simpler ORM usage, otherwise LGTM.

PR #1090, opened 2019-07-15
Add screenshot route

jcushman:
Just a little Monday-funday -- this is a take on letting our /trends/ links show previews of the graph, and would also allow us to do cool stuff in the future like having ?highlight= links show an image of the highlighted text.

The approach is to have a python route that calls a node program, which loads the live URL and returns a screenshot of a given CSS selector using puppeteer. Dunno if this is a good idea, but it seems cool. Deployment: requires node_modules to be installed on production. Currently behind a SCREENSHOT_FEATURE flag.


bensteinberg (approved):
:+1:

PR #1091, opened 2019-07-18
Volume and reporter model updates

jcushman:
Draft for now; suggest waiting to pull this until after the data release, so as not to slow that down.

A bunch of stuff preparatory to fixing up our nominative reporter cites:

- Reporter and VolumeMetadata models get history tracking
- Add EditLog and EditLogTransaction models that can aggregate a set of changes to history-tracked models under a text description
- New `duplicate` and `duplicate_of` fields to track that a volume or case is a duplicate of another
  - Previous `volume.duplicate` field is renamed to `legacy_duplicate`. `case.duplicative` stays around -- it's confusingly similar but means something different.
- New `task_statuses` JSON field on volumes; refactor fab tasks to use this field. This gives us a generic way to re-run fab tasks that run per-volume, only if they have not succeeded after a certain date. It also records the error message for volumes that fail.
- New `CaseMetadata.in_scope` field on cases, that aggregates the logic for whether a case should show in our public data.
  - Drop old partial indexes and add new ones that use `in_scope`, using the new Django 2.2 partial index support.
  - New `fab update_in_scope` task to set the value of the field for all volumes. (Don't run this now; see below)
- New helper methods to change a volume's reporter or duplicate status and update the appropriate fields on its cases.

I expect I've missed some edge cases that will need to be updated, since there's a lot of moving parts here ... 🤷‍♂️

DEPLOYMENT:

For this initial deployment, it will be faster to NOT use the new fab task. Instead, we can do this as a one-off after applying migrations: `update capdb_casemetadata set in_scope=false where duplicative=true or jurisdiction_id is null or court_id is null;`.


jcushman:
Whoops, turning into a draft.

PR #1093, opened 2019-07-19
Gallery cms

ChefAndy:
New capweb database
New gallery model in capweb/models
Modified admin to work with new gallery data
Modified gallery view to process new cms data
Modified gallery template accept new cms data
Modified gallery CSS


README.md
```diff
@@ -129,6 +129,7 @@ This will make a virtualenv entitled "capstone." You can tell that you're inside
 
     (capstone)$ psql -c "CREATE DATABASE capdb;"
     (capstone)$ psql -c "CREATE DATABASE capapi;"
+    (capstone)$ psql -c "CREATE DATABASE capweb;"
```
jcushman: What do we get out of adding a third database here? I think the new models want to live in the capapi database, unless we get something out of adding a third deploy strategy.

capstone/capweb/migrations/0003_auto_20190712_1929.py
```diff
@@ -0,0 +1,74 @@
+# Generated by Django 2.2.3 on 2019-07-12 19:29
+
+from django.db import migrations
+
+def load_gallery_initial_dataset(apps, schema_editor):
```
jcushman: In general we don't want to put fixtures in migrations, because they have different semantics -- for example, pytest runs migrations only once but recreates fixtures it needs for each test. For these you could do something like `./manage.py dumpdata --indent=2 capweb.Gallery > capweb/fixtures/gallery.json` and then add them to `fab ingest_fixtures` (though I wouldn't look into doing this until we figure out the image and section questions below).

capstone/capweb/models.py
```diff
@@ -1 +1,15 @@
 # Create your models here.
+from django.db import models
+
+
+class Gallery(models.Model):
+    title = models.CharField(max_length=255)
+    section = models.CharField(max_length=20, default="Applications")
```
jcushman: I think it's probably worth having a GallerySection model with name and order fields, and a foreign key here. This setup is going to make it tricky to do things later like have a page per section.

capstone/static/css/scss/gallery.scss
```diff
@@ -1,5 +1,6 @@
 @import 'variables';
 @import 'footer/magenta';
+@import 'base.scss';
```
jcushman, line 3: If we import base.scss, `{% block extra_head %}` has to become `{% block base_css %}` in gallery.html to avoid double css in the page.

capstone/capweb/models.py
```diff
@@ -1 +1,15 @@
...
+    content = models.TextField() # markdown-formatted text
+    image = models.ImageField(blank=True, null=True, upload_to="static/img/gallery/")
```
jcushman, line 15: This isn't going to work on production; the static/ folder is only served live with `DEBUG = True`, and newly uploaded files won't be copied over to staticfiles/. Likewise `{% static entry.image %}` isn't going to work in the template.

This is actually pretty tricky to do correctly. For starters it should look something like this:

- set `MEDIA_ROOT = os.path.join(BASE_DIR, 'media')` in settings_base
- set `MEDIA_URL = '/media/'`
- add this bit to our urls.py so the `/media/` URL works in development: https://docs.djangoproject.com/en/2.1/howto/static-files/#serving-files-uploaded-by-a-user-during-development
- update nginx to serve BASE_DIR + 'media' at the `/media/` URL in production

Then you can do `{{ entry.image.url }}` in the template to link to it.

The downside of just using nginx to serve the media folder in production is we can't scale to multiple servers, and that folder on the web server ends up holding data that isn't backed up anywhere. Ideally production would use django-storages and a remote storage, or an NFS mount, or a custom database-backed storage. We should chat with @bensteinberg about whether we feel like doing this correctly right now -- I think maybe not.

jcushman (changes requested):
Good start! Some suggested changes ...

ChefAndy:
re-opening as a draft PR

PR #1097, opened 2019-07-23
Highlight case text

anastasia:
This PR adds highlighting any text included in the `?highlight=` parameter

Keeping it to a single (short) JS file instead of moving everything to Vue.js for now to keep case.html light. What do you think?

Screen Shot 2019-07-23 at 1 44 39 PM


jcushman (changes requested):
I'm good with not dropping in Vue.js at this point -- nbd to migrate over if we need it. A couple of issues, though -- 

If I'm reading it right, this approach won't work with cross-paragraph highlights or highlights that include italics, which will be common. One option I see for a library that would work is https://markjs.io/ with `"acrossElements": true`, though there may be better options.

I also don't think it'll work to treat the query param as a regex -- there's too many random punctuation characters that might end up in highlight params. Maybe at some point we could add a feature where `?highlight=/foo/` is treated as a regex, tho ...

PR #1098, opened 2019-07-23
Gallery cms

ChefAndy:


ChefAndy:
* ~~New capweb database~~
     * Removed new database and started using default (capapi) database
* New in capweb/models
     * New GalleryEntry model which includes entries on the gallery page
     * New GallerySection model, which groups GalleryEntry models into sections
     * New model to store CMS images
     * Storage which stores the data in the image fields and stores them in the CMS image model
          * The latest few versions of django-storages don't seem to support database as a back-end storage. The django-db-file-storage seesm to do the trick well enough, though it seems to be a bit wonky. I don't think it would be particularly difficult to switch to some other system, either now or later. I'd like to get @jcushman's thoughts on sticking with this. 
* Added test to make sure new entries storage works
* Modified admin to work with new gallery data
* Modified gallery view/template to accept new cms data
* Modified gallery CSS

PR #1100, opened 2019-07-26
 Add pdf rendering celery task

jcushman:
To generate all PDFs (on AWS): `fab make_pdfs`
To generate one PDF: `fab make_pdfs:redacted/32044032310971_redacted`

By default, will only generate PDFs that don't already exist.

DEPLOYMENT:
- requires `libtiff-tools` and `pdftk` packages (standard apt-get, see dockerfile)
- requires `settings.STORAGES['pdf_storage']` to point wherever we want PDFs to go

Next step I think is to try generating a few on AWS and confirm that they look how we want.


jcushman:
After optimizing, I think this task runs at about 25 pages per second, so say 3 or 4 weeks of CPU time.

PR #1123, opened 2019-08-07
Highlight case text 2

anastasia:
This PR is the beginning of getting a URL on highlighting functionality. There are a few outstanding issues I'm dealing with, but in the spirit of continuous integration, pushing this out right now because it's already taken waaay too long!

Issues needing to be resolved:
- on selecting large paragraphs and clicking copy link, sometimes (though not always??) window scrolls down some unpredictable amount. I'm _assuming_ this is an edge case because this feature seems most useful for citation/phrase highlighting
- using screen reader with this feature: on Becky's recommendations, we're adding the tooltip element (an `<a>` tag) right after the text being selected (as a sibling element). I need to make sure that this element is in fact select-able with a screen reader.
- not sure if this is an issue? Tooltips are fixed positions (we need them to float above the selected text and accomplish that with some hackiness) which means that if someone selects text, gets the tooltip, and scrolls any way on the page, the tooltip will stick to the viewport. All someone has to do is click off to deselect the text and therefore hide the tooltip. This might be annoying.


anastasia:
added a couple weirdness fixes:
- tooltip should be set to the center every time (sometimes it appeared off to the left if multiple HTML elements were selected because the width ended up being too big)
- auto-created bootstrap tooltip div now added to the DOM in a screenreader friendly manner as well

PR #1132, opened 2019-08-14
Images

anastasia:
getting images from cases
saving images to storage


capstone/fabfile.py
```diff
@@ -1181,4 +1181,23 @@ def update_in_scope(last_run_before=None):
 # this is convenient for profiling, e.g. "kernprof -l fabfile.py refresh_case_body_cache"
 if __name__ == "__main__":
     from fabric.main import main
-    main()
\ No newline at end of file
+    main()
+    main()
```
bensteinberg: Is this right, calling `main()` twice?

anastasia: it needed to be sure!

PR #1139, opened 2019-08-22
Another error in the API documentation - cite vs citation

GPPassos:
Small fix in API documentation again, compare

https://api.case.law/v1/cases/?cite=409 U.S. 188 (works, count = 1)
to
https://api.case.law/v1/cases/?citation=409%20U.S.%20188 (does not work)


jcushman:
Thanks!

PR #1141, opened 2019-08-23
Mailchimp signup

ChefAndy:
Killed our internal mailing list sign-up functionality and pointed our mailing list sign-up form right at our mailchimp campaign.


capstone/capapi/migrations/0016_delete_mailinglist.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 2.2.4 on 2019-08-23 18:38
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
jcushman: Let's wait to delete the model until the rest of this lands, so we don't have to be too precise about the timing of copying the data over.

capstone/capweb/templates/index.html
```diff
@@ -609,10 +614,15 @@ <h2 class="section-title pt-0 mb-3">
         </div>
         <div class="d-none d-md-block col-5 col-md-5 col-lg-3 offset-1 offset-sm-0">
 
-          <form class="email_form_large" action="{% url 'subscribe' %}" method="post">
-                {% bootstrap_form subscribe_form %}
-                <div class="add">Add your email to keep in touch</div>
-                <button class="btn-secondary subscribe" type="submit">Subscribe!</button>
+          <form action="https://law.us3.list-manage.com/subscribe/post?u=4290964398813d739f2398db0&amp;id=e097736c6f"
```
jcushman: Can we make the magic numbers here (and same elsewhere) into settings? That way dev/staging sites won't post to the real mailing list.

jcushman (changes requested):
Thanks! Just a couple of tweaks.

capstone/capweb/templates/index.html
ChefAndy: Sure.

capstone/capapi/migrations/0016_delete_mailinglist.py
ChefAndy: Ok.

jcushman:
Thanks!

Issue #1152, opened 2019-08-28
Expose API call on search results page


abziegler:

ChefAndy:
Merged in #1165

PR #1165, opened 2019-09-05
API Query Explainer

ChefAndy:
Little educational feature in the search page to expose the API URL a la #1152, and to highlight the corresponding parts of the query url based on moused-over/focused input fields.


jcushman (changes requested):
Total nitpick -- we use 2 spaces for indents in JS and HTML, vs. 4 in Python. Can you tweak that?

ChefAndy:
> Total nitpick -- we use 2 spaces for indents in JS and HTML, vs. 4 in Python. Can you tweak that?

Done!

PR #1167, opened 2019-09-11
Adds filters box to elasticsearch case endpoint

ChefAndy:
Overrode to_html method in CAPFiltering class (which extends FilteringFilterBackend) which creates renders the filter form in html.

Added a template for the filter form.


capstone/capapi/views/api_views.py
```diff
@@ -146,6 +147,112 @@ def get_filter_query_params(self, request, view):
...
+
+        # this is for the content of the dropdown menus for jurisdiction, court, and reporter
+        options = {}
+        options['jurisdiction'] = [ {
+                'value': jurisdiction.slug,
+                'label': jurisdiction.name_long,
+                'selected': 'selected' if 'jurisdiction' in filter_values
+                                          and jurisdiction.slug in filter_values['jurisdiction'] else ''
+                }
+            for jurisdiction in models.Jurisdiction.objects.all()]
```
jcushman: Oh, this is what you were asking about the other day! Rather than load these from the DB each time, they can be imported as jur_choices, court_choices, and reporter_choices from filters.py.

jcushman (changes requested):
Nice! One tweak ...

capstone/capapi/views/api_views.py
ChefAndy: Done!

PR #1189, opened 2019-09-26
Skip out-of-scope volumes

### LINKED ISSUE #1005: Case browser 500 error for reporters with blank volume numbers
E.g. https://cite.capapi.org/ga-app/

`django.urls.exceptions.NoReverseMatch: Reverse for 'volume' with keyword arguments '{'volume_number': '', 'series_slug': 'ga-app'}' not found. 1 pattern(s) tried: ['(?P<series_slug>[^/]+)\\/(?P<volume_number>[^/]+)\\/$']`

bensteinberg:
I'm not sure this is right or idiomatic.

Fixes #1005 ?

### DIFF
--- a/capstone/cite/views.py
+++ b/capstone/cite/views.py
@@ -77,11 +77,11 @@ def series(request, series_slug):
         return HttpResponseRedirect(helpers.reverse('series', args=[slugify(series_slug)], host='cite'))
     reporters = list(Reporter.objects
         .filter(short_name_slug=series_slug)
-        .prefetch_related(Prefetch('volumes', queryset=VolumeMetadata.objects.exclude(volume_number=None).exclude(duplicate=True)))
+        .prefetch_related(Prefetch('volumes', queryset=VolumeMetadata.objects.exclude(volume_number=None).exclude(volume_number='').exclude(duplicate=True).exclude(out_of_scope=True)))
         .order_by('full_name'))
     if not reporters:
         raise Http404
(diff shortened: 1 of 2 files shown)


bensteinberg:
(However, I just tried this as a hotfix on beta and it does work.)

capstone/cite/views.py
```diff
@@ -81,7 +81,7 @@ def series(request, series_slug):
         .order_by('full_name'))
     if not reporters:
         raise Http404
-    reporters = [(reporter, sorted(reporter.volumes.all(), key=lambda volume: natural_sort_key(volume.volume_number))) for reporter in reporters]
+    reporters = [(reporter, sorted(reporter.volumes.all().exclude(out_of_scope=True), key=lambda volume: natural_sort_key(volume.volume_number))) for reporter in reporters]
```
jcushman: You want to put this .exclude() up in line 80 so it doesn't result in a separate sql query. The `prefetch_related` is making sure that these `reporter.volumes.all()` calls won't actually result in a new database hit, but only if they're unfiltered -- hence the change in `django_assert_num_queries`. (Thanks, tests!)

Looking at this again, prefetch_related is totally a premature optimization -- there will only ever be one or two reporters anyway, so using prefetch_related only reduces our query count by like one and I shouldn't have bothered. It would also be valid to move all the filters down to line 84 and get rid of the prefetch_related for the sake of simplicity, and bump django_assert_num_queries to 3. 🤷‍♂️

bensteinberg: Ah, OK, thanks. I just pushed another test "correction", but will go back and do it this way.

PR #1211, opened 2019-10-08
Simple es fixtures & test updates

ChefAndy:
### fixtures.py, *json
* temporary data and fixtures to be updated when fixtures based on token stream ingest are implemented.

### documents.py, settings_base.py, settings_dev.py:
 * Changed index based on whether testing or not.

### serializers.py
 * Court was giving the ID in the url instead of the slug.

### api_views.py
 * In some tests parameters were passed to the filters box without values, so the dictionary key for 'value' wasn't created. This was not a problem outside of test but with --fail-on-template-vars it got tripped.

### fabfile.py
* rebuild_search_index now has a force option to surpass the confirmation dialog during testing


bensteinberg:
Sorry about all the churn here, but the tests are now running with Elasticsearch on Circle CI. They're failing like this there
```
capapi/tests/test_api.py FF........F......FF....F..........
```
and being inconsistent locally over different test runs:
```
capapi/tests/test_api.py FF........................^C ### too slow, i gave up
capapi/tests/test_api.py .......................F...........
capapi/tests/test_api.py ..................F....F...........
capapi/tests/test_api.py ..........F......FF....F...........
capapi/tests/test_api.py ..........F............F...........
capapi/tests/test_api.py ...................................  ###  passed!!!
capapi/tests/test_api.py ..................F................
```

PR #1216, opened 2019-10-18
Unpin pip, upgrade pip-tools

bensteinberg:


capstone/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN apt-get update \
 RUN mkdir /app
 WORKDIR /app
 COPY requirements.txt /app
-RUN pip install pip==19.2.3 \
+RUN pip install -U pip \
```
jcushman: Not sure that we want to unpin pip here rather than updating the number manually -- the intention is for every docker instance across dev instances and CI to have the same versions of everything.

bensteinberg: Sure -- this is what we had before f7c76dd, FWIW. I'll pin it to 19.3.1.

PR #1217, opened 2019-10-21
Mailing list signup

ChefAndy:
Adds a checkbox to the register screen asking users if they'd like to sign up for Lawvocado. Upon saving the new user, if they selected the checkbox, it will add them to Lawvocado in Mailchimp with a 'pending' status so they'll receive a confirmation email. The alternative to this would be to save their selection in the database and then check to see if they checked the checkbox when they confirm their email account with case.law. I'm leaning towards this because it feels a little more opt-in-ish.
**The settings.MAILCHIMP['api_user'] and settings.MAILCHIMP['api_key'] settings will need to be filled in for production use.**


bensteinberg:
I'm not sure it absolutely matters, but you might rebase on develop again and rerun `docker-compose exec web fab pip-compile` to keep the docker image numbers going in ascending order.

PR #1220, opened 2019-10-24
Context menu

anastasia:
This PR adds functionality: it moves the Copy URL tooltip to a context menu of sorts, adds copy text, search Google, search DDG buttons. It fixes the problem of dragging and dropping to a new tab to automatically search selected text.

gif

This code is the result of hitting my head against the wall for many days trying to make this feature accessible. Here are some things that were tried:
- In order for an element to be tab-able, it needs to be in the correct order for the user to tab to. That means that the context menu needs to be dynamically moved around and placed as the closest possible sibling of the text that is selected.
In order to achieve that, a no-brainer is to simply lift the menu and append it as sibling: `selection.focusNode.parentNode.appendChild(contextMenu);` No matter what shenanigans I pulled with this context menu, I wasn't able to get the event handlers to work on the buttons.
-- **The problem**: while this works flawlessly in Firefox, on Chrome and Safari the events tied to the `<a> tags` in the menu, somehow, are either not created at all or become unbound. This means that although the menu gets placed in the correct position regardless of the browser, the actions on the menu in Chrome and Safari are rendered useless. I thought at first that this was an event clobbering problem, but more and more it seems that the event handlers don't even get successfully created.
-- **The hypothesis**: This could be a browser bug, though as yet I have not been able to find valid similar issues (there are some out there that point to different code problems). The problem in certain browsers could also be related to assigning event handlers on an element that's out of view(port). In the beginning, I'm keeping the menu to the left of the window, so that it's hidden until we need it. Maybe there are rules about what can be clickable?
- Thanks to @rebeccacremona for this one: by assigning a `tabindex` attribute on the fly, we're able to control what element the user tabs to next. 
-- **The problem**: tabbing away from selected text deselects it. OK, that's a concession that we can maybe make (the highlighted becomes deselected, the menu pops open with all the selected text options, but since we've saved the user's selection it's only visually jarring). In Firefox, we're able to programmatically assign tabindexes and, thanks to pretty ugly code (where we're making sure that we're focusing on the right things and not collapsing the menu when tabbing through it) this works. In Chrome, while it looks like the tabindexes are assigned correctly, the order is not respected. The user's tab jumps from an expected place, back to body, as opposed to into the context menu.
-- **The hypothesis**: Maybe this is a problem having to do with the `display` property? I haven't found anything about this yet, but could it be that the menu is not having the `.show()` function applied quickly enough for the `tabindex` order to work effectively?


anastasia:
Updates!
- it looks like `appendChild` may have not been the problem on Chrome, rather, as I thought at first, it was event clobbering. Moving from using `mouseup` to `onselectchange` allowed the `click` events of the context menu to be heard (note: do not trust Chrome dev tools' Event Listeners tab! It seems to be very moody in what events it includes). 
- Using a debouncer for `onselectchange` (keeping the increment small or else copying text doesn't work)
- This will need to be tested in Chrome, and probably fixed up. My browser has not been great for that: the recommended "enable caret" extension for keyboard browsing doesn't work, and I haven't been able to find a substitute. Currently Chrome is picking up on the hidden-from-view input tag (used for executing the copy command for `copyURL` and `copyText`). Hiding it from view (type="hidden", or display: none) disables the copy feature. I've put the input field at the end of the DOM so that should solve it, but I haven't been able to test this because of the caret browsing problems. 
Still needs work: 
- I need to install `skiplinkfix` so that when tabbing backwards, the user can find their way back to the context menu (it's currently out of view)
- Something funny is going on where context menu gets hidden to quickly so the success text is barely noticeable.

rebeccacremona:
Congrats on all the progress! Quick note: I'm not confident the `skiplinkfix` will help with getting the screen to scroll back to the menu when using shift+tab. I'll have a think on what might.....

jcushman:
Cool! Let's try this on beta.

PR #1221, opened 2019-10-25
add - hot reload

anastasia:
I know there are varying opinions on this in the group — we should talk about this!
Currently, yarn server reloads, inconsistently, on frontend changes. This PR makes the reload consistent in development.

### DIFF
--- a/capstone/vue.config.js
+++ b/capstone/vue.config.js
@@ -63,7 +63,11 @@ let vueConfig = {
     headers: { 'Access-Control-Allow-Origin': '*' },
     allowedHosts: [
         '.case.test'
-    ]
+    ],
+    watchOptions: {
+      poll: true
+    }
+
   },
 
   chainWebpack: config => {


jcushman:
I like live updates, but can you double check what kind of CPU hit you see when enabling this option? What this PR does is to use manual polling to check for updates instead of relying on filesystem events, which makes it more likely to work in places like Docker file mounts that might not support filesystem events properly. It's not enabled by default because manual polling is much more CPU intensive -- though in practice that might be negligible if it's only polling a small directory in the first place.

anastasia:
So anecdotally, it does take a bigger hit, like 10% more per CPU core. However, again anecdotally, running this with a bunch of other CPU heavy apps open (everything from several browsers to CPU-heavy adobe applications), I haven't noticed a slowdown yet. 

With hot reload:
Screen Shot 2019-10-30 at 9 59 42 AM

Without hot reload:
Screen Shot 2019-10-30 at 10 21 24 AM

jcushman:
That seems ok to me!

I poked around a bit and it looks like increasing CPU by 10 percentage points would maybe use like 4 watts more power, which means that, say, your computer might run out of battery half an hour sooner with this setting on. I think that's probably OK in exchange for more reliable debugging?

PR #1222, opened 2019-10-28
add - exposing "file" in root level of downloads. file includes all paths in /download

anastasia:
Screen Shot 2019-10-28 at 3 51 41 PM

### DIFF
--- a/capstone/capapi/templates/file_download.html
+++ b/capstone/capapi/templates/file_download.html
@@ -25,13 +25,6 @@
     </nav>
   {% endif %}
   <ul class="list-group">
-    {% if isroot %}
-      <li class="list-item">
-        <a href="{% url 'download-contents' %}">
-          manifest.csv
-        </a>
-      </li>
-    {% endif %}
     {% for file in files %}
       <li class="list-item">
         {% if file.isdir %}
@@ -44,7 +37,9 @@
           {% else %}
             {{ file.name }}
           {% endif %}
-          <span class="file-size small color-gray">{{ file.size | filesizeformat }}</span>
+          {% if file.size %}
+            <span class="file-size small color-gray">{{ file.size | filesizeformat }}</span>
+          {% endif %}
         {% endif %}
       </li>
 
--- a/capstone/capapi/templates/file_download.html
+++ b/capstone/capapi/templates/file_download.html
@@ -24,6 +24,11 @@
     </ol>
   </nav>
   {% endif %}
+  {% if isroot %}
+      <a class="dl-list-of-contents" href="{% url 'download-contents' %}">
+        Download a full list of contents
+      </a><br/><br/>
+  {% endif %}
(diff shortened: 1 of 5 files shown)


capstone/capweb/urls.py
```diff
@@ -48,6 +48,7 @@
     ### downloads ###
     path('download', views.download_files, name='download-files'),
     re_path(r'download/(?P<filepath>.*)', views.download_files, name='download-files-in-path'),
+    path('download_contents', views.download_contents_file, name='download-contents'),
```
jcushman: Let's put this above the previous line, with the path "download/manifest.csv", so it works more like a real file.

capstone/capweb/views.py
```diff
@@ -373,3 +375,17 @@ def download_files(request, filepath=""):
             "error": "This file was not found in our system."
         }
         return render(request, "file_download_400.html", context, status=404)
+
+
+def download_contents_file(request, filepath=""):
+    absolute_path = download_files_storage.path(filepath)
+    with tempfile.TemporaryFile() as tf:
```
jcushman: Rather than writing to a temp file, I would use the csv writer here with StringIO to avoid encoding problems of building your own string or the need to write out to disk: https://stackoverflow.com/a/9157370/307769 

For the csv contents, let's add column headers and have three columns for the path, file size, and modification date. This file will be cached so I think the cpu hit should be ok.

capstone/capapi/templates/file_download.html
```diff
@@ -10,21 +10,28 @@
...
+            <li class="breadcrumb-item">
+              <a href="{% url 'download-files' %}/{{ breadcrumb.path }}">{{ breadcrumb.name }}</a>
+            </li>
+          {% endif %}
+        {% endfor %}
+      </ol>
+    </nav>
   {% endif %}
   <ul class="list-group">
+    {% if isroot %}
```
jcushman: It's a little weird to have this item not sorted alphabetically. It might actually simplify the view code a little to refactor a bit so manifest.csv doesn't need special treatment in the template: have the view put together a list of dicts like `[{"name":, "path":, "is_dir":, "size":}]`, sort by name, and then the view doesn't need a special file wrapper and the template doesn't need to know that the manifest.csv dict is special.

jcushman (changes requested):
Thanks! I suggested a few more changes ...

capstone/capapi/templates/file_download.html
anastasia: But the manifest.csv will be a special case still since it won't be generated until called by the user, right? So we won't know how large the file is?

jcushman: Yep, true! I think that's fine ...

jcushman:
Thanks for taking a second pass at that -- looks great!

PR #1227, opened 2019-11-04
Make cite.case.law use ES for cases instead of DB

ChefAndy:
## capstone/capapi/documents.py
* changed the data type of some fields
* added some other fields to reporters embedded object
* started to pull in some functionality from the CaseMetadata model like get_frontent_url and full_cite

## capstone/capapi/serializers.py
* changed the way the get_volume function gets the volume

## capstone/cite/templates/cite/citation_failed.html
* changed the reference from first to [0] for citations

## capstone/cite/templates/cite/series.html
* updated variable names to work with new data types

## capstone/cite/views.py
*  updated most places that touched the database to use ES instead. Some exceptions include querying non-cases tables. 

## capstone/scripts/refactor_xml.py 
* when writing token_streams the volume number was being put in the reporter dictionary but wasn't being read in when creating the volume. It now reads it in when creating the volume.

## capstone/test_data/test_fixtures/fixtures.py
* added some CaseDocument fixtures.
* removed old dict fixtures

## tests
* added cases back to the endpoint test
* removed the test ingest because the endpoint test does the same thing
* updated to work with the new views
* updated to use new fixtures


capstone/capapi/documents.py
```diff
@@ -96,11 +110,38 @@ class Meta:
...
+        auto_refresh = False
+
+    def full_cite(self):
+        return "%s, %s%s" % (
+            self.name_abbreviation,
+            ", ".join(cite.cite for cite in self.citations),
+            " (%s)" % self.decision_date.year if self.decision_date else ""
+        )
+
+    def get_frontend_url(self, disambiguate=False, include_host=True):
```
jcushman: Does this function ever get called? I don't think it can work. The general context here is that it's hard to generate frontend URLs, because they depend on whether the case shares a citation with any other case. So we (should) only use `CaseMetadata.get_frontend_url()` in situations where we've efficiently figured out whether `disambiguate` should be true, and are using the result to set `CaseMetadata.frontend_url`. Otherwise we don't know if `disambiguate` should be True or False, and we should just use `CaseMetadata.frontend_url`.

ChefAndy: It works in that it creates the URL, but it was only used in the cite/test_views tests and never got called with the disambiguate flag. I copied it over from CaseMetadata to CaseDocument because it was used in the cite tests, and figured we'd probably start moving some of the functionality over anyway so I'd get the ball rolling.  Easy enough to take out and only use frontend_url!

jcushman: Yeah, I'd say pull it out since it can't really return correct results and I don't want it to be misleading. Could make a simpler test helper if faking something for tests is helpful.

ChefAndy: Done

PR #1228, opened 2019-11-05
Draft README for image download folder

bensteinberg:


bensteinberg:
I just changed the name of the folder from `images` to `illustrations`; I kind of like `figures` better, but I think `illustrations` will be more legible to many readers.

PR #1233, opened 2019-11-18
Move admin URLs above download URLs so gallery image URLs work

bensteinberg:


bensteinberg:
Thanks for the feedback, @jcushman. I've corrected the regex but left the admin URLs where I'd moved them -- I don't think the concatenation of that section in https://github.com/harvard-lil/capstone/blob/develop/capstone/capweb/urls.py#L82-L89 is necessary?

PR #1237, opened 2019-11-19
Add ability to temporarily include cases in robots.txt

jcushman:
With a hat-tip to @mlissner in https://github.com/freelawproject/courtlistener/issues/1007 , this allows us to temporarily add cases to cite.case.law/robots.txt, which will hopefully cause crawlers to immediately pick up changes to a given case's noindex meta tags.

Also tweaks the case admin page to refresh elasticsearch on save.


mlissner:
Glad to see that bug caught your eye. I'm still unclear of whether it's really an improvement. I still get people complaining that their case isn't down, but it's possible I fulfilled their request before the new robots.txt thing was in place. 

I'll be curious to hear if this works for you guys. How long are things staying on your robots.txt before falling off the bottom?

jcushman:
Dunno, I was just putting this substrate in place, and then we can write `now() + 7 days` or whatever to the datetime field as we go along. I suppose we should check how often robots.txt is being hit in the logs and pick a time window from there.

mlissner:
Wow, I don't know why I didn't think of that. I'm an idiot sometimes. 

I think ours is set to seven days, IIRC. Erring on the side of too long seemed like the right approach, so that's where it landed. Our file is HUGE though, so that's the countervailing argument. We're at about 8000 entries right now. :/

PR #1253, opened 2019-12-11
Redacting or eliding text in case

anastasia:
Screen Shot 2019-12-11 at 1 45 46 PM


capstone/capdb/models.py
```diff
@@ -915,6 +921,14 @@ class Meta:
     def save(self, *args, **kwargs):
         if self.in_scope != self.get_in_scope():
             self.in_scope = not self.in_scope
+
+        # Add a custom footer message to instance if redactions or elisions exist but no text is provided
+        if not self.custom_footer_message and (self.no_index_redacted or self.no_index_elided):
```
jcushman, line 926: I would do this in the view so it's easier to change later.

jcushman (commented):
Nice work! One suggested change ...

capstone/capdb/models.py
anastasia: If we do this in the view, we're going to have to do another case lookup. I'm trying to keep case lookups to a minimum.... Do you think that's needless to worry about?

jcushman:
Let's try this out on beta while we continue to chat about that one nit ...

PR #1265, opened 2019-12-18
Volume slug

ChefAndy:
## documents.py 
* added volume_number_slug field in volume object

## models.py 
* VolumeMetadata
    * added volume_number_slug field
    * added function to update it
    * overrode save function so it updates automatically
* CaseMetadata
    * get_frontend_url
        * changed regex that matches the citation pattern to see if disambiguation is needed. In group two, it now checks for a capital letter as the first character, because the '1/2' in '81 1/2' was tripping it up.
        * have it use the volume_number_slug rather than volume_number

## tasks.py
* made a update_volume_number_slug task which updates the slugs and frontend_urls in both the DB and elasticsearch

## test_models, test_tasks, cite/test_views
* updated test logic

## citation_failed.html, series.html, volume.html, urls.py, views.py
* updated to use volume_number_slug

## volume_metadata.json
*  updated test metadata to include volume_number_slug

## fab
* made update_volume_number_slugs task


bensteinberg:
I think you may need to bump the migration number, as @anastasia got 0085 already: https://github.com/harvard-lil/capstone/pull/1253/files#diff-050faf17e4c4825ac0f69777c88125c1

capstone/capdb/tasks.py
```diff
@@ -115,6 +116,27 @@ def sync_from_initial_metadata_for_vol(self, volume_id, force):
...
+        # if the volume number and volume number slug aren't the same, the cases' frontend_url needs to be updated.
+        # Else, just, just reindex the cases as they are to give ES the new slug value
+        if vol.volume_number != new_slug:
+            # this performs a re-index on the elasticsearch CaseDocuments so there's no need to update the slug manually
+            CaseMetadata.update_frontend_urls([ case.citations.first().cite for case in vol.case_metadatas.all() ])
+            print("Changing {} to {} ({}) required modifying {} cases for {}".format(
+                original_slug, vol.volume_number, vol.volume_number_slug, vol.case_metadatas.count(), barcode))
+        else:
+            print("Updated {} Model and ES with new slug value.".format(barcode))
+            CaseMetadata.reindex_cases([ case for case in vol.case_metadatas.all() ])
```
jcushman: This is much simpler than the implementation of `update_elasticsearch_for_vol` -- shouldn't we call that task instead? And then I think you would delete the `prefetch_related("case_metadatas__citations")` above.

capstone/capdb/models.py
```diff
@@ -979,17 +987,23 @@ def get_frontend_url(self, cite=None, disambiguate=False, include_host=True):
...
             # try to match "(year)-(series)-(case index)", e.g. "2017-Ohio-5699" and "2015-NMCA-053"
             m = re.match(r'(\S+)-(.+?)-(\S+)$', cite.cite)
         # TODO: final fallback value is wrong, because first_page is the physical page count and not the page label
         # after token streams are in, we should be able to retrieve the actual page label instead
-        cite_parts = m.groups() if m else [self.volume.volume_number, self.reporter.short_name, self.first_page]
-        args = [slugify(cite_parts[1]), cite_parts[0], cite_parts[2]] + ([self.id] if disambiguate else [])
+        volume_number, rep_short_nm, fp = m.groups() if m else [self.volume.volume_number_slug, self.reporter.short_name, self.first_page]
+
+        if self.volume.volume_number_slug:
+            volume_number = self.volume.volume_number_slug
```
jcushman: Maybe cleaner just to use `slugify(volume_number)` below instead of using self.volume.volume_number_slug? This current code assumes that the citation we're using to build the frontend_url will always match the book volume, which I *think* is true but maybe isn't necessary to assume.

capstone/capdb/models.py
```diff
@@ -979,17 +987,23 @@ def get_frontend_url(self, cite=None, disambiguate=False, include_host=True):
             If disambiguate is true, return /series/volnum/pagenum/id/.
         """
         cite = cite or self.citations.first()
-        # try to match "(volnumber) (series) (pagenumber)"
-        m = re.match(r'(\S+)\s+(.+?)\s+(\S+)$', cite.cite)
+
+        # try to match "(volnumber) (Series) (pagenumber)"
+        m = re.match(r'^(\S+)\s+([A-Z].+?)\s+(\S+)$', cite.cite)
```
jcushman: Is it worth adding special handling for the `81 1/2` and `33 Suppl.` cases here? The regex could look like:

`m = re.match(r'^(\S+(?: Suppl\.| 1/2)?)\s+([A-Z].+?)\s+(\S+)$', cite.cite)`

This way we don't end up setting `disambiguate = True` for those two volumes unnecessarily.

jcushman (commented):
Cool! I'm just trying to get my brain going again after break -- let me know if these comments make sense.

capstone/capdb/models.py
ChefAndy: Done

capstone/capdb/models.py
ChefAndy: Done

capstone/capdb/tasks.py
ChefAndy: Done

PR #1285, opened 2020-01-13
Download with command line

anastasia:
sending back lists of files if request looks like a command line request (https://github.com/harvard-lil/capstone/projects/36#card-27572757)

### DIFF
--- a/capstone/capapi/templates/file_download.html
+++ b/capstone/capapi/templates/file_download.html
@@ -24,7 +24,7 @@
       </ol>
     </nav>
   {% endif %}
-  <ul class="list-group">
+  <ul class="list-group download-files-list">
     {% for file in files %}
       <li class="list-item">
         {% if file.isdir %}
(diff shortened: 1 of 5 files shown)


jcushman:
I like this approach! Two suggested changes:

(1) Let's hook into DRF's content negotiation logic instead of using custom logic for `is_browser_request`, so these pages behave the same as our API pages:

```
from rest_framework.negotiation import DefaultContentNegotiation
from rest_framework import renderers
from rest_framework.request import Request as RestRequest

def is_browser_request(request):
    """
    Differentiate between command line and browser requests
    """
    drf_renderer = DefaultContentNegotiation().select_renderer(RestRequest(request), [renderers.JSONRenderer, renderers.TemplateHTMLRenderer])
    return drf_renderer[0] is renderers.TemplateHTMLRenderer
```

(2) I think it's confusing for manifest.csv to return JSON. Let's add a second pseudofile at manifest.json instead.

PR #1286, opened 2020-01-13
add recovery key to cases endpoint

jcushman:
The primary goal of this PR is to add `"recovery_key"` to the cases endpoint.

In adding tests for that, I also brought some test helpers and refactoring over from h2o:

- new `as_user` parameter for client.get()
- replace pytest-factoryboy with a simpler and more reliable `@register` decorator
- refactor user and client fixtures

### DIFF
--- a/capstone/capapi/models.py
+++ b/capstone/capapi/models.py
@@ -137,7 +137,7 @@ class CapUser(PermissionsMixin, AbstractBaseUser):
 
     def authenticate_user(self, activation_nonce):
         if self.activation_nonce == activation_nonce and self.nonce_expires + timedelta(hours=24) > timezone.now():
-            Token.objects.create(user=self)
+            Token.objects.get_or_create(user=self)
             self.activation_nonce = ''
             self.email_verified = True
             self.save()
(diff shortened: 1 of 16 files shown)


bensteinberg:
Should the salt be a setting?

jcushman:
I don't think it needs to be a setting? The function of the salt here is as a namespace -- it ensures that recovery_key signatures can't be derived from other signatures, because `Signer(salt="recovery_key").sign('1-1')` won't match `Signer().sign('1-1')` even if Django somehow exposed other values from Signer elsewhere in the app. It wouldn't make sense to ever change this within a given deployment, because that would break existing keys, and there's no need for it to be different in different environments.

Another reason you might want a setting is just on the principle that if you're going to use a string like `"recovery_key"` in two places in code, you should assign it to a constant like `RECOVERY_KEY_SALT = "recovery_key"` somewhere in your app to make the fact that it *is* a constant more explicit and type-checkable, and `settings` is the easiest place to put that. I've semi-guiltily chosen to use bare strings directly in various places recently instead of defining constants, because I'm not convinced that the layer of indirection buys you anything at our relatively-small scale -- you need roughly the same degree of testing and rigor to ensure that your constants are being used correctly as that your string literals are being used correctly, but you're using more code to do it. Dunno, could be talked out of that.

jcushman:
The bigger question I'm having is, are we sure we want to commit to pushing this work onto the clients in the name of privacy, or can we find a way to track this on the backend that we're happy with.

PR #1289, opened 2020-01-15
enable doctests

jcushman:
Pytest will now test all code examples in docstrings.


bensteinberg:
Tests at Circle took ~5 minutes longer than, for instance, #1285 just now.

jcushman:
I think that's misleading -- the running-python-tests phase at Circle was actually faster for this PR than that one.

(I've been wondering about test speed overall, though -- I think we can probably pick up some speed if we decide to work on that.)

PR #1303, opened 2020-01-22
Case histories

jcushman:
This adds a user setting to track cases accessed. The primary impact of tracking user history is that non-open cases will not be counted against the user's quota if accessed again in the future.

Here's how it looks -- the setting before opt-in:

image

The "what does this mean" panel:

image

The setting once selected:

image

The history page:

image

There is also an API endpoint at /v1/user_history/ that allows you to query your history by case_id and date range.

**Deployment notes**

All history data is stored in a separate database, allowing us to have separate (shorter) retention periods for backups.

Local devs will need to run `docker-compose exec db psql --user=postgres -c "CREATE DATABASE cap_user_data;"` and `fab migrate` to create the new database.

Production deployments will need a new database set up with an appropriate backup schedule.

**Also tucked into this PR**

As usual a crop of quick fixes are coming along for the ride, as earlier commits so they can be kept if we have to roll back the main thing here:

* Fix run_task_for_volumes date filter
* Drop a redundant serializer
* Tweak the display of the response headers in our API html browser. Before:
image
After:
image
* Fix cite pagination in PDF export (citations weren't necessarily ending up on the correct pages before)
* Move an incorrectly placed target=blank on court links in the search page


bensteinberg:
Let's not merge this until after today's data release and deployment.

PR #1305, opened 2020-01-22
update - allowing clickable elisions in case title

anastasia:


jcushman:
Thanks! Do the spans need to be excluded elsewhere in the template, like `"description": "{{ metadata.name }}"` ?

PR #1327, opened 2020-02-03
update - directing user to previous page after login

### LINKED ISSUE #1077: Login link doesn't take you back where you were
Steps to reproduce:

 - Go to a page requiring login while logged out.
 - Click the login link in upper right
 - Log in.

Results:

 - You're now at the homepage.

Expected results:

 - You're back at the place where you clicked the link

We figured this out on freelawproject/courtlistener a while back. I think the standard Django login view handles a `redirect_to` parameter, so if you just have a `?redirect_to={{ some_url }}` in your template, I think you can just pass that to the view. 

You can find our code in our base.html template, but the only trick I remember is that you have to make sure you avoid a redirect loop from the homepage...that can get ugly IIRC.

anastasia:
fixes #1077

### DIFF
--- a/capstone/capapi/templates/registration/login.html
+++ b/capstone/capapi/templates/registration/login.html
@@ -25,7 +25,7 @@
             class="btn btn-primary">
       Log in
     </button>
-    <input type="hidden" name="next" value="{{ next }}" />
+    <input type="hidden" name="next" value="{{ request.META.HTTP_REFERER }}" />
   </form>
   <p>
     Don't have an account?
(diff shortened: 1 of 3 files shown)


capstone/capapi/templates/registration/login.html
```diff
@@ -25,7 +25,7 @@
             class="btn btn-primary">
       Log in
     </button>
-    <input type="hidden" name="next" value="{{ next }}" />
+    <input type="hidden" name="next" value="{{ request.META.HTTP_REFERER }}" />
```
jcushman: This won't work if the user enters the wrong password the first time, or if the login form is linked from off-site. Maybe a better fix for the open issue would be to add `?next={{ request.get_full_path }}` to the login link in the header?

anastasia: ACK, great catch!!

PR #1367, opened 2020-02-19
Update fixtures

jcushman:
The primary goal here is to stop using the ingest fixtures (`ingest_django_fixtures` and `ingest_elasticsearch`) which load json objects from disk, and instead only use factoryboy fixtures. This makes the fixtures considerably faster, and also easier to configure for specific tests.

Also in this PR:

* Other speed improvements for tests: reenable `pytest --nomigrations` and disable whitenoise middleware
* Monkeypatch elasticsearch library, and modify CaseMetadata save function, so there's no need to explicitly reindex after save or to retry elasticsearch queries in tests
* Fix tokenization of case api filters so name searches with punctuation in them will work (I know this seems unrelated, but it was necessary to get a test to pass)
* Delete some scripts and tests for them that were based on modifying case xml instead of token streams


jcushman:
On test speed: this reduced the "Test" phase on CircleCI from 13m 56s in the most recent PR before this, to 5m 23s for this one. On my local machine the tests will run in less than two minutes.

PR #1368, opened 2020-02-20
Redactions -> Body Cache

ChefAndy:
sync_case_body_cache now redacts text in no_index_redacted. (The body cache feeds ES.)


bensteinberg:
Does this PR mean that redactions will require a data release including ES index?

jcushman:
Hmm, let's talk about it. It might be fine for redactions to only hit the raw API/bulk when we do full releases every quarter or so; they're also applied during rendering of the case viewer, so we don't need a data release for that. If we do want them to hit the raw API right away, we could move this logic to the renderer.

bensteinberg:
OK, thanks, TBD.

PR #1371, opened 2020-02-21
Add regexes

bensteinberg:
Thanks for the regexes, @jcushman. 

What I want here is something like a list comprehension for `Q() | ...`

### DIFF
--- a/capstone/capdb/tasks.py
+++ b/capstone/capdb/tasks.py
@@ -58,21 +58,34 @@ def record_task_status_for_volume(task, volume_id):
 ### TASKS ###
 
 @shared_task(bind=True, acks_late=True)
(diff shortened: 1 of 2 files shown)


bensteinberg:
Ah, maybe `reduce(operator.or_,`, will try it.

bensteinberg:
Thanks for the edit -- I thought the A should get an X, too?

jcushman:
Aw dang. Yeah I think it actually wants to be "AXX-XXXXXXXXX" -- the A is analogous to the dashes.

bensteinberg:
Ah, OK. I guess, like with SSNs, we are not concealing that this is a masked A number. I'll put in another PR unless you're already working on one.

jcushman:
Go for it -- thanks!

PR #1374, opened 2020-02-21
State-level access logging

jcushman:
Friday Funday feature to add a log message like "Someone from Massachusetts read a case from Supreme Court of the United States." when a case is accessed in the case browser.

Requires `GEOLOCATION_FEATURE = True` in settings, and the file `test_data/GeoLite2-City.mmdb`, which I retrieved from https://dev.maxmind.com/geoip/geoip2/geolite2/ and can pass along.


jcushman:
(Actually give me a minute -- let's catch errors from this so it can't hurt anything if the IP address is missing.)

jcushman:
OK carry on.

PR #1375, opened 2020-02-21
Run some tests in parallel with pytest -n 2

jcushman:
Experimental -- if this works on Circle, it means that tests that don't access the db don't have to wait for those that do.


jcushman:
This speeds up tests for me by 30% locally, but only 5% on Circle because it has less CPUs and more overhead. Oh well, I guess worth keeping around?

PR #1380, opened 2020-02-25
Case pdfs

jcushman:
Now that we have some volume PDFs in /download/, we can add links to the individual cases to view a PDF for that case.

image

PDF links look like `/pdf/<case_id>/<full_cite>.pdf`, e.g. `http://cite.case.test:8000/pdf/5/Cavazos v. Smith, 565 U.S. 1 (2011)`

Cases with redactions currently return a 404 for their PDF, and cases will not have a PDF link if we haven't posted the volume PDF yet (i.e. currently this only works for open jurisdictions, because those are all we've downloaded).

This PR also adds the full-volume-pdfs directory `/download/PDFs/` to robots.txt to prevent search engine indexing.

**Deployment:** This requires `apt-get install -y ghostscript poppler-utils` on the beta/prod servers, as in the Dockerfile.


bensteinberg:
Did you look at PyMuPDF? I used that for another project and thought it was both good and fast.

jcushman:
I didn't -- I'll benchmark it.

jcushman:
PyMuPDF is definitely simpler -- it's just `pip install PyMuPDF` and then

```
        import fitz
        doc = fitz.open(self.volume.pdf_file.path)
        doc.select(range(page_range['order__min']-1, page_range['order__max']))
        doc.save('/app/temp.pdf')
```

... but the resulting PDF is 50MB instead of 800KB. I think I'm missing a step to fully remove the other pages, or else it's uncompressing the images or something. Can look into it more later ...

jcushman:
OK, with an extra `garbage=2` PyMuPDF works great -- something like 10 times as fast as my previous version, and no apt-get requirements. Thanks for the suggestion!

(I noticed the tests were stalling out, and added a `-v` to the pytest command on circle to see what test was stalling, and that one ran fine. Hopefully just a random one-off glitch.)

bensteinberg:
Deployment to beta is failing -- there does not appear to be a wheel, so I need to "download and generate MuPDF": https://github.com/pymupdf/PyMuPDF#installation

Stay tuned.

bensteinberg:
Upgrading pip to 20.2 allowed `pip install pymupdf` to install the wheel, but then `pip-sync` was broken; after downgrading back to 9.0.3, `pip-sync` was still broken, but upgrading to 19.3 appears to have done the trick.

PR #1390, opened 2020-02-27
Extract citations

anastasia:
extracting citations, writing them to DB

mostly: 
but also: this is a work in progress PR. Some mysterious stuff is happening where new citations are being written each time I run this fab task. I'm 90% sure it's not the regex (how ...could it be?). It's gotta be something to do with DB caching.


anastasia:
It's looking like our es CaseDocument search is serving different results on each run. More runs = more complete results, but this shouldn't be happening to begin with. Investigating

mlissner:
Somebody flagged this for me. I promise I'm not watching all the work you guys do! But I thought I'd just Sea Lion my way in here to say it's exciting to see the reporters_db hard at work.

2014-09-19-1062sea

 Let me know if you have questions. We're actively working on both the reporters_db and our citation regexes/system, so it's all on our minds.

anastasia:
Still need to sort out where to write citation misses (OCR errors, just plain mistakes picked up by the regex, etc). Also, would love to talk to you @mlissner about your regex!

mlissner:
@anastasia if you email me your email address, I'll get you in our slack. I'm mike@free.law.

anastasia:
This PR is lacking a place for the csv of the citation misses! Any thoughts @jcushman, @bensteinberg, @ChefAndy? We will want to revisit the misses at some point in time because they could include fixable OCR errors.

mlissner:
> This PR is lacking a place for the csv of the citation misses!

FWIW, we set up our citation finder so it can just be re-run whenever we have fixes. It's a pain, but the fixes are fairly infrequent. In any case, it avoids the need for a CSV.

PR #1391, opened 2020-02-27
Remove format=html, move logic to case browser view function

jcushman:
* Remove `format=html` from cases API endpoint and docs
* Move and simplify logic from HTMLRenderer to case browser's view function
* Make redaction in CaseDocument apply to name_abbreviation, to harmonize with current frontend redaction
* Make frontend redaction use same helper function as CaseDocument

### DIFF
--- a/capstone/capapi/documents.py
+++ b/capstone/capapi/documents.py
@@ -1,6 +1,7 @@
 from django_elasticsearch_dsl import Document as DocType, Index, fields
 from django.conf import settings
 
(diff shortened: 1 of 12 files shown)


jcushman:
Pushed a tweak that redirects requests for `api.case.law/v1/cases/ID/?format=html` to the frontend case browser.

jcushman:
Update to also remove XMLRenderer, per conversation in Slack.

PR #1397, opened 2020-03-03
Load cites from ES instead of DB

jcushman:
Whoops, forgot (again!) that we don't have the citations table populated on prod.


jcushman:
Hold on this -- one more tweak coming.

jcushman:
NM we'll fix this by copying the citations table over.

PR #1411, opened 2020-03-16
Extract citations

anastasia:
- add normalized_cite filed
- getting only in_scope cases
- map/reducing missed citation results


capstone/fabfile.py
```diff
@@ -1147,6 +1148,17 @@ def extract_all_citations(last_run_before=None):
...
+    def concat_and_remove_file(smallfile):
+        with open("/tmp/missed_citations.csv", "a+") as extracted_citations_file:
+            writer = csv.writer(extracted_citations_file)
+            with open(smallfile) as f:
+                reader = csv.reader(f)
+                [writer.writerow(row) for row in reader]
+            os.remove(smallfile)
+
+    [concat_and_remove_file(f) for f in glob('/tmp/missed_citations-*.csv')]
+
```
jcushman: This doesn't make sense to do here, as it will run before any of the volumes are processed. Maybe we just do it as `cat /tmp/missed_citations-*.csv >> missed_citations.csv` after the tasks are complete?

capstone/capdb/tasks.py
```diff
@@ -430,13 +432,14 @@ def extract_citations_per_vol(self, volume_id):
...
+            normalized_cite=normalize_cite(c["cite_original"]),
             cited_by=c["cited_by"],
             reporter_name_original=c["reporter_name_original"],
             volume_number_original=c["volume_number_original"],
             page_number_original=c["page_number_original"]) for c in extracted_citations])
 
-    fieldnames = ['case_origin', 'missed_cites_per_case', 'missed_cites']
-    with open("/tmp/missed_citations.csv", "a+") as f:
-        writer = csv.DictWriter(f, fieldnames=fieldnames)
+    with open("/tmp/missed_citations-%s.csv" % self.request.id, "w+") as f:
```
jcushman: Let's put this in a subfolder like `/tmp/missed_citations` in case anyone needs to find anything in the /tmp dir while it's running. Can throw in a line like `Path("/tmp/missed_citations").mkdir(exist_ok=True)` first.

capstone/capdb/models.py
```diff
@@ -1662,10 +1663,15 @@ class ExtractedCitation(models.Model):
     reporter_name_original = models.CharField(max_length=200)
     volume_number_original = models.CharField(max_length=64, blank=True, null=True)
     page_number_original = models.SmallIntegerField(null=True, blank=True)
+    normalized_cite = models.SlugField(max_length=10000, null=True, db_index=True)
```
jcushman, line 1666: I think we drop the db_index on `cite_original` when adding this, to make inserts faster?

jcushman (changes requested):
Cool! Just a few tweaks.

Issue #1412, opened 2020-03-18
rendering of mailto in "403 - Access to this file is restricted" page


bensteinberg:
When trying to visit a restricted PDF, logged in but not as a researcher, the error page renders a mailto link incorrectly in one place but correctly in another - roughly:

    403 - Access to this file is restricted
    If you believe you should have access to this file, please let us know at <a href='mailto:info@case.law'>info@case.law</a>.
    If you got here through our site or an article we host, please let us know at info@case.law so we can fix it.

(The second email address is a link.)

PR #1413, opened 2020-03-19
unautoescaped output on error_page

ChefAndy:
Addresses: #1412

### DIFF
--- a/capstone/capweb/templates/error_page.html
+++ b/capstone/capweb/templates/error_page.html
@@ -10,6 +10,7 @@
 {% block top_section_style %}bg-black{% endblock %}
 
 {% block explainer %}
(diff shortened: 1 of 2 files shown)


jcushman:
This needs a little more digging -- error page output is weird in Django because of a series of security bugs involving injecting JS into custom error responses. Want to make sure we're not undoing that.

bensteinberg:
I think you still have debugging code at line 303... ;)

ChefAndy:
> I think you still have debugging code at line 303... ;)

Whoops— missed a commit in that push.

ChefAndy:
We'll try this later.

ChefAndy:
Hmm... worked locally. Let's try that again.

PR #1426, opened 2020-03-31
Gallery Sections

ChefAndy:
Full implementation of gallery sections. On the galleries page, this will only show entries marked as "featured." At the bottom of each list, is a link to view the rest of the entries on a separate page. On that page, which is similar in layout to the galleries page, all entries are listed.

I also created the 'exhibits' url "directory"/namespace for witchcraft, wordclouds, and limericks. It seems that the gallery section has functionality that is distinct enough from our hosting static self-made content that we should have a different place for those things. Now, the official URL for witchcraft, for example, will be ```https://case.law/exhibits/witchcraft```, and I think that all new static features like that should be placed in the exhibits directory/namespace. The old urls for those three exhibits will redirect with a 302.

The only difference required in managing gallery entries and exhibits is deciding whether or not an entry should be "featured" on the homepage. That can be adjusted at any time in the admin interface.

* capstone/capweb/views.py
    * added the gallery section view
    * filtered out non-featured items from the main gallery view
* capstone/capweb/urls.py
    * added the gallery section urls
    * added the exhibits urls
* capstone/capweb/templates/gallery_section.html 
    * the template for the gallery section view
* capstone/capweb/templates/gallery.html
    * added the links to the gallery section pages
* capstone/capweb/models.py
    * Added the title_slug field to the GalleryEntry and GallerySection models
    * Added the 'featured' field to the GalleryEntry model
* capstone/capweb/migrations/0003_auto_20200331_1845.py
    * title_slug fields
* capstone/capweb/migrations/0002_galleryentry_featured.py
    * featured flag
* capstone/capweb/admin.py
    * added the featured field to the admin view


anastasia:
This is awesome, looks great!

PR #1433, opened 2020-04-03
Cite extraction tweaks

jcushman:
Some improvements to citation extraction:

* Don't match citations where volume or page number starts with `0`
* Match citations where volume number ends with `Suppl.` or `1/2`
* Don't extract citations where reporter name starts with a digit (no valid reporter names do)
* Remove footnote numbers from CaseBodyCache.text
* Filter out invalid reporters like `or`
* Translate OCR-error reporters like `la.`
* Update Elasticsearch as we go along, so we don't have to rebuild the search index afterward

Also included:

* prefetch_related citations in update_elasticsearch_for_vol
* drop one-off task `sync_xml_image_case_body_cache_for_vol`
* `case` fixture is now a built-in `CaseFactory` instance; fix tests impacted by that
* add test for `sync_case_body_cache`
* `docker-compose.yml` wanted to be updated for some reason; maybe two mutually-incompatible PRs we brought in recently?

### DIFF
--- a/capstone/capapi/resources.py
+++ b/capstone/capapi/resources.py
@@ -22,7 +22,7 @@ from capapi.tasks import cache_query_count
 from capweb.helpers import reverse, statement_timeout, StatementTimeout
 from config.logging import logger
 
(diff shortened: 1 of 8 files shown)


jcushman:
Pulling to run tests locally, since CircleCI is throwing what looks like a temporary SSL error building the database image.

PR #1440, opened 2020-04-07
exporting citation connections

anastasia:
Writing to disparate json files per vol (like with our extraction citations), adding to one big json file using a fab task.


jcushman (changes requested):
Just for documentation of Slack conversation -- Anastasia is working on a second iteration of this that handles citations resolving to more than one case.

anastasia:
Thanks @holtchesley for pointing to `json_agg`

PR #1444, opened 2020-04-09
Copy: About page reflects case pdfs

holtchesley:


holtchesley:
This is more of a note about the about info being updatable than an assertion about the quality of the copy I wrote (which I am in no way attached to).

jcushman:
Thanks!

PR #1450, opened 2020-04-16
Retry emails if first attempt to send fails.

rebeccacremona:
@bensteinberg says CAP is experiencing the same sporadic email errors that Perma was, e.g.:
```
Internal Server Error: /user/register/
    
SMTPServerDisconnected at /user/register/
Connection unexpectedly closed
```

I applied this patch in Perma, which simply makes another attempt to send the email a second later, if the first attempt fails. We haven't seen any exceptions since.


rebeccacremona:
Note: I didn't test locally... my Docker build failed while compiling rocksdb.

jcushman:
Thanks!

PR #1451, opened 2020-04-17
Citation dir creation and filetype errors

anastasia:


jcushman:
Thanks! I rolled back the part of this that moved from the csv writer to raw string processing, since the csv format turned out not to be the problem with opening the gzip on Mac.

PR #1453, opened 2020-04-21
add - citation graph readme

anastasia:
a short readme (including stolen content form @jcushman). Does it need to be more extensive? Things to add, potentially: how to get more data about each case knowing its case id? Different graphing libraries? A visualization?


jcushman:
Thanks!

PR #1465, opened 2020-05-12
Build and commit JS at Circle CI

bensteinberg:


.circleci/config.yml
```diff
@@ -69,12 +69,21 @@ jobs:
...
+      # Commit built assets if necessary, then deploy via Salt reactor
       - run:
           name: "Deploy"
           command: |
-            if [ "$CIRCLE_PULL_REQUEST" == "" -a "$CIRCLE_BRANCH" == "develop" ] ;
-            then
+            if [[ "$CIRCLE_PULL_REQUEST" == "" && "$CIRCLE_BRANCH" == "develop" ]] ; then
+                if [[ `git status static/dist/ webpack-stats.json --porcelain` ]] ; then
+                    docker-compose up -d
+                    docker-compose exec web yarn build
```
jcushman: `yarn build` has to go somewhere before the `git status` check, right? In h2o it goes in the js test block, which makes sense, except we don't have js tests. Could go in a this-is-where-js-tests-would-hypothetically-go block.

PR #1472, opened 2020-05-22
Dup vol remediation

ChefAndy:
Added a few simple model fields, added some model functions to populate those and other fields, added a script to make some duplicate volume entry modifications, and created some tests.

I removed all of the PDF renaming/deleting stuff I was working on because I just wasn't confident in the way it would all work out in production. I'll put that through in a different pull request along with a script that modifies the existing pdf entries.


capstone/capdb/models.py
```diff
@@ -633,32 +649,65 @@ def set_xml_checksums_need_update(self, value=True):
...
             self.duplicate_of = duplicate_of
             self.save()
-            self.case_metadatas.update(duplicate=is_duplicate)
+            self.case_metadatas.update(duplicate=True)
             self.case_metadatas.update_in_scope()
+        CaseMetadata.reindex_cases(self.case_metadatas.all(), prune_duplicates=True)
 
-    def set_reporter(self, reporter):
+    def set_volume_number(self, volume_number, update_citations=True):
+        """
```
jcushman, line 667: Nit: update comment to refer to this function.

capstone/capdb/models.py
```diff
@@ -970,9 +1033,26 @@ def update_frontend_urls(cls, cite_strs, update_elasticsearch=True, batch_size=1
...
     @classmethod
-    def reindex_cases(cls, cases):
+    def reindex_cases(cls, cases, prune_duplicates=False):
         from capapi.documents import CaseDocument  # avoid circular import
-        CaseDocument().update(cases)
+        if isinstance(cases, models.Model):
+            cases = [cases]
+
+        # only indexes non-duplicate cases
+        CaseDocument().update([case for case in cases if not case.duplicate])
```
jcushman: should use in_scope instead of duplicate?

capstone/capdb/models.py
```diff
@@ -970,9 +1033,26 @@ def update_frontend_urls(cls, cite_strs, update_elasticsearch=True, batch_size=1
                     cls.reindex_cases(to_update)
 
     @classmethod
-    def reindex_cases(cls, cases):
+    def reindex_cases(cls, cases, prune_duplicates=False):
```
jcushman: Having `reindex_cases` remove out of scope cases is a great idea! Can we make it happen all the time, and depend on case.in_scope instead of on case.duplicate?

capstone/capdb/models.py
```diff
@@ -24,7 +24,8 @@
 from django.utils.encoding import force_bytes, force_str
 from django.core.files.base import ContentFile
 
-from capdb.storages import bulk_export_storage, case_image_storage, download_files_storage, writeable_download_files_storage, pdf_storage
```
jcushman, line 27: Seems like this file had a code formatter run on it? NBD, but since we don't currently have a formatter standardized for the project, I'd rather not run random formatters that generate a lot of arbitrary changes in the diff.

jcushman (commented):
Awesome work! Just a couple of tweaks ...

capstone/capdb/models.py
ChefAndy: Whoops— normally I only format the code I write. I must have accidentally run it without any code selected.

capstone/capdb/models.py
ChefAndy: Yep.

capstone/capdb/models.py
ChefAndy: Do'h.

capstone/capdb/models.py
ChefAndy: Yep.

PR #1491, opened 2020-06-12
Case editor -> working towards a first pass

ChefAndy:
For some reason many of the cache tests aren't passing. I need to figure that out.


ChefAndy:
So they only fail locally, it seems. I'm blowing away my docker containers to see what the deal is. Until then, I'm making this a regular pull request.

PR #1502, opened 2020-07-06
Add tools sidebar to single case view

jcushman:
image

This is a redesign of the single-case page -- we'll surely keep redesigning from here!

Accomplishes a few things:

* Replace the selection context menu with a "selection tools" sidebar, as the context menu was interfering with ordinary text selection and copying, in different ways that were hard to get right across browsers.
* Add a case outline linking to each opinion in the case.
* Add more room to explain the API and PDF format options.
* Add room for an explanation of what the document even is, for readers arriving from Google.

Todo:

* Once this is on beta I'd love feedback on the keyboard flow, mobile layout, phrasing, general design etc.


jcushman:
Oh another goal was to add space for linking to other tools like a list of cases citing this case.

PR #1511, opened 2020-07-09
Tweaked case viewer styling

ChefAndy:
That is all.


ChefAndy:
I did a quick refresh of the case viewer styling to better visually incorporate the sidebar. The way the name_abbreviation sits at the top gives the page an unbalanced feel. It's a pretty heavily-used component of the website, so it could probably use a deep-dive rethinking at some point, but it looks a bit more put-together now.

PR #1512, opened 2020-07-10
Case editor updates

jcushman:
Taking a pass at Andy's case editor ...

* Send the entire token stream for processing in JS, instead of pre-processing in Python.
* Send back a list of edits which get saved to the database. Edits are passed around as [oldVal, newVal], so we can detect and decide what to do if there's attempted simultaneous editing of the same data.
* Try displaying OCR pages as `` and `<span>` tags instead of via canvas. This lets us use Vue for reactivity and CSS for styling, which seems to work out pretty well.
* Send font information to JS and use it for OCR display and alignment, including adjusting vertical alignment based on character ascents.
* Save state to localStorage until it is saved to database.


ChefAndy:
💯

PR #1536, opened 2020-07-29
Track updated cases in CaseLastUpdate and ES index in background

jcushman:
* Drop the CaseMetadata.last_update field and replace with a CaseLastUpdate one-to-one table
* Add postgres triggers so any change to a field used by the API will update the CaseLastUpdate table
* Add celerybeat task to reindex any cases in CaseLastUpdate that haven't been indexed yet, once per minute
* Make `fab run` start celerybeat
* Remove everywhere else in the code that we care about Elasticsearch indexing

Also in this PR:

* Remove some extra ES fields we don't use
* Change citation extractor so it only saves changed citations


jcushman:
After this is in, I have a SQL query to populate the new table.

PR #1616, opened 2020-10-22
Use our own registry

bensteinberg:


bensteinberg:
From https://github.com/harvard-lil/perma/pull/2864 --
> It is redundant and possibly confusing to replicate all the necessary images from Docker Hub to our own registry, but it may save us from rate-limiting by IP in CI, and having our own registry may be useful in other contexts.

PR #1622, opened 2020-10-27
Separate documents for ES reading and writing

jcushman:
This allows the Elasticsearch reading and writing to be separated, so the index can be rebuilt while the site is still serving results from the old index. The process would look something like this:

* set indexes read only
* use ES clone API to make _ro copies
* set original indexes back to read-write
* point Django to the _ro copies:
```
ELASTICSEARCH_INDEXES={
    'cases_endpoint': 'cases',
    'resolve_endpoint': 'resolve',
    # these can be set on prod if they're different:
    'cases_reader_endpoint': 'cases_ro',
    'resolve_reader_endpoint': 'resolve_ro',
}
```
* run `fab rebuild_search_index`. This will delete `cases` and `resolve` endpoints, but we'll keep serving results from the clones.
* once reindexing is done, point Django back to the originals and delete the _ro indexes.

### DIFF
--- a/capstone/capapi/documents.py
+++ b/capstone/capapi/documents.py
@@ -9,18 +9,17 @@ from capdb.models import CaseMetadata, CaseLastUpdate
 from scripts.simhash import get_distance
 
 
-def get_index(index_name):
-    index = Index(settings.ELASTICSEARCH_INDEXES[index_name])
+## helpers
+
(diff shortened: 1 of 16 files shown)


bensteinberg:
Closing this for the moment -- it may not be necessary.

PR #1623, opened 2020-10-28
Docs Reorg

ChefAndy:
Big docs reorg.

Kicked some of the "general" items back into the regular page nav. 

Split things up into "User Pathways" which only includes the guide for researchers right now, "Specs and Ref" which is supposed to be more terse technical reference material that can be quickly consulted, "User Guides" which are supposed to be more friendly, prosey kind of documentation, and then tutorials which are step-by-step teaching tools to teach conceptual basics. Now contains a CAP API tutorial.


jcushman:
Closing in favor of #1633 that includes these commits

PR #1642, opened 2020-12-07
Search download2

anastasia:


anastasia:
New search page

Screen Shot 2021-01-02 at 9 34 09 PM

Advanced filters (show/hide)

Screen Shot 2021-01-02 at 9 34 47 PM

Results shown (on large screen)

Screen Shot 2021-01-02 at 9 35 10 PM

Download options

Screen Shot 2021-01-02 at 9 35 28 PM

Filters are dismiss-able if used

Screen Shot 2021-01-02 at 9 35 48 PM

Still to do:
x some CSS cleanup is still needed like padding and margin inconsistencies on various window sizes (but it has driven me to the brink and I am taking a break)
- loading icon when first loading search is not in an ideal place
- downloading options div needs some love
- making sure that everything is accessible (dropdowns and other fields should be)
- icons need hover/active/focus states but my new laptop doesn't have sketch and won't download illustrator because of some mysterious permissions problem
- have to figure out the off-by-one download limit (currently capped at 9,999 instead of our 10,000)

PR #1670, opened 2021-01-07
group head matter tags

ChefAndy:
Groups most head matter tags into divs, and adds styling. 
Also changes some other styling to make it look a bit more cohesive.

### DIFF
--- a/capstone/cite/tests/test_views.py
+++ b/capstone/cite/tests/test_views.py
@@ -120,7 +120,16 @@ def test_single_case(client, auth_client, token_auth_client, case_factory, elast
 
     # set up for viewing html or pdf
     case_text = "Case HTML"
-    unrestricted_case = case_factory(jurisdiction__whitelisted=True, body_cache__html=case_text, first_page_order=2, last_page_order=2)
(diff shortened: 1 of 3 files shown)


jcushman (changes requested):
I like the look of this redesign, but the code currently reorders paragraphs in the head matter in my testing. For example see the attorney tags in https://cite.case.law/us/5/309/. (That volume can be loaded locally with `fab import_web_volumes` if you don't already have it).

image

Maybe we could get the same effect without reordering the html at all? You can style the first (but not last) paragraph of a kind with css like ```:not(.summary) + .summary {
    border: 1px red solid;
}```, so if you use that to insert the label and top margin and use `.summary { }` to set the background color maybe that gets you there. I admit that does allow less flexibility in the design tho.

One other suggestion is to style the other headmatter that isn't currently grouped, for example giving it the same font size and margins, so it's more clear that all the headmatter goes together. For the messier cases it's going to be a little confusing to have normally formatted text interspersed with small gray text -- it sort of suggests the unstyled headmatter is part of the case text.

ChefAndy:
Putting in a new PR

PR #1676, opened 2021-01-13
Update Postgres in dev

### LINKED ISSUE #1674: Update dev postgres to 11.10
We're running Postgres 11.10 on prod. Let's bump this dockerfile from 9.6 for dev.

https://github.com/harvard-lil/capstone/blob/develop/services/docker/extended-postgres.dockerfile

bensteinberg:
This updates Postgres to 11.10 in our Docker development environment. There are a few points of interest:

To get your environment set up again, run

    docker-compose up -d
    docker-compose exec db psql --user=postgres -c "CREATE DATABASE capdb;"
    docker-compose exec db psql --user=postgres -c "CREATE DATABASE capapi;"
    docker-compose exec db psql --user=postgres -c "CREATE DATABASE cap_user_data;"
    docker-compose exec web fab init_dev_db
    docker-compose exec web fab ingest_fixtures
    docker-compose exec web fab import_web_volumes

In order to make a clean and easy break with the previous setup, this changes the names of both the database and index volumes. You can keep the old ones around for reference, or delete them with

    docker volume rm db_data
    docker volume rm esdata01

Note the new method for installing the deprecated `temporal_tables` extension in `services/docker/extended-postgres.dockerfile`. The pinned commit in the checkout line is the same as checking out
https://github.com/mlt/temporal_tables/tree/mlt -- see https://github.com/mlt/temporal_tables for a recommendation to use
https://github.com/xocolatl/periods instead.

Finally, I had to increase the memory allotted in Docker Desktop's Preferences... Resources to prevent the Elasticsearch container from dying with code 137 -- I set it to 4GB. Your mileage may vary. It's not clear to me whether the added use of memory was inherent in this change, or in inessential excess load I was putting on the index while I was fussing with it.

Closes #1674.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ Capstone should now be running at 127.0.0.1:8000.
 
 ### Docker Setup <a id="docker-setup"></a>
 
-We have initial support for local development via `docker compose`. Docker setup looks like this:
+We support local development via `docker compose`. Docker setup looks like this:
 
     $ docker-compose up -d
     $ docker-compose exec db psql --user=postgres -c "CREATE DATABASE capdb;"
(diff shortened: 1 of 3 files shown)


bensteinberg:
(Replacing the ES volume may not be necessary, but I think can't hurt; I did that before figuring out the out-of-memory error.)

PR #1717, opened 2021-02-22
Correctly label chronological order of "cited by" page

raindrum:
If an opinion is cited by more than one page's worth of opinions, they are split across multiple pages and the button to go to the second page is currently labeled "older entries." See [example](https://cite.case.law/cited-by/2292816/). That's backwards, since the citations are listed from oldest to newest. This PR swaps the "older entries" and "newer entries" labels to reflect that.

An alternative fix would be to keep the labels as-is, but reverse the order of cases.


jcushman:
Thanks!

PR #1737, opened 2021-03-08
Labs branding

ChefAndy:
The title bar is a little hackey, but it shows the labs logo and the name of the project. Hamburger menu doesn't work. Each labs project will need a project_name codeblock from now on.

Also added stub text for the about Chronolawgic page.


ChefAndy:
Merging because @anastasia's 2FA wasn't working

PR #1754, opened 2021-03-22
Update some packages

bensteinberg:
I was hoping to get the JS packages to upgrade without adding them to `package.json`, but a simple `yarn upgrade` didn't do it (and upgraded a lot of other stuff) -- what I ran for 281d730 is

    docker-compose exec web yarn upgrade 'is-svg@~>4.2.2'
    docker-compose exec web yarn upgrade 'ssri@~>8.0.1'
    docker-compose exec web fab update_docker_image_version


bensteinberg:
Looks like the DRF update will require some changes.

jcushman:
After googling around it looks like the correct way to set a minimum for sub-dependencies is to add them to "resolutions":

```
  "resolutions": {
    "vue-svg-loader/svg-to-vue": "0.6.0",
    "**/is-svg": "^4.2.2",
    "**/ssri": "^8.0.1"
  }
```

... and then run plain `yarn` to regenerate yarn.lock.

https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-why-would-you-want-to-do-this

This way those packages won't keep being installed if we drop whatever dependency requires them.

bensteinberg:
See https://www.django-rest-framework.org/community/release-notes/#3112 for the change from `urlize_quoted_links` to `urlize`. 

I don't understand why I had to remove `.decode()` in https://github.com/harvard-lil/capstone/pull/1754/files#diff-914b15229d485d468e64a278af8f98bbf1ccb1cf60bfb1226963b29988652870L18

PR #1760, opened 2021-03-24
[ chrono ] basically functional mobile vue- modal magic needs potion modified

ChefAndy:


bensteinberg:
I think this may be missing `static/img/icons/more-horizontal.svg`

PR #1763, opened 2021-03-25
Upgrade pip and pip-tools

jcushman:
I ran into an issue installing a recent python package from source, which is fixed by upgrading pip-tools. So ...

* Update pip from 19.3.1 to 21.0.1
* Update pip-tools from 4.2.0 to 6.0.1 (this causes a lot of noise in requirements.txt because of some formatting changes)


jcushman:
For deployment, this may require updating pip separately on stage/prod since that's handled outside of requirements.txt.

PR #1764, opened 2021-03-25
Use force_str so we don't have to care if text is bytes

bensteinberg:


bensteinberg:
Thanks for the tip, @jcushman.

PR #1769, opened 2021-03-30
Let CI manage Docker image version

bensteinberg:


.circleci/config.yml
```diff
@@ -76,12 +76,13 @@ jobs:
...
                 docker-compose exec web yarn build
-                if [[ `git status static/dist/ webpack-stats.json --porcelain` ]] ; then
-                    docker-compose exec web fab update_docker_image_version
+                docker-compose exec web fab update_docker_image_version
+                if [[ `git status static/dist/ webpack-stats.json docker-compose.yml --porcelain` ]] ; then
                     git config user.email "lil@law.harvard.edu"
                     git config user.name "Circle CI"
                     git add static/dist/ webpack-stats.json docker-compose.yml
-                    git commit -m 'Add built JS [skip ci]'
+                    JSCHANGED=`git diff --name-only --cached | grep -q stat && echo ', add built JS'`
```
jcushman: If you want the commit message to be precise, note that both are optional -- `git status static/dist/ webpack-stats.json` implies JS was rebuilt, and `git status docker-compose.yml` implies docker version was updated.

jcushman: You could also break them out entirely as yarn build, commit changes to static/dist/ webpack-stats.json if any; fab update_docker_image_version, commit changes to docker-compose.yml if any.

bensteinberg: Ah, right. I was thinking that if we were in this stanza, we'd by definition have a new Docker image number, but that's wrong. Hold one, I'll push a fix.

bensteinberg: Well.... if we've updated JS, we should have a new image number, and if we haven't updated JS and don't have a new image number, we're not going to be making a commit here. Am I wrong?

PR #1771, opened 2021-03-30
[chrono] UUID

ChefAndy:
Swaps out the numerical DB id as the user-facing timeline identifier.


bensteinberg:
On deployment to stage, when `collectstatic` runs, we get an error:
```
                          File "/app/labs/models.py", line 1, in <module>
                            import shortuuid
                        ModuleNotFoundError: No module named 'shortuuid'
```
I think this needed a bump of the image number in `docker-compose.yml` in order to force a container rebuild, and I may need to add something to #1769 to handle requirements changes.

jcushman:
I think #1769 should already cover it, it just wasn't in yet. The `update_docker_image_version` call will update the docker image version if requirements change.

PR #1778, opened 2021-04-01
Ensure ES server is up

bensteinberg:
This should eliminate a class of sporadic connection errors when running tests at Circle CI. I can reliably produce this kind of connection error with `docker-compose down ; docker-compose up -d ; docker-compose exec web pytest` -- the ES container is not available by the time it's needed.


bensteinberg:
Thanks again, @jcushman!

PR #1784, opened 2021-04-06
[chrono] basic clobber stopper and test

ChefAndy:
Was going to split up all of the timeline operations into separate API endpoints, but since the whole point is to avoid clobbering the existing timeline if the site gets out-of-sync, this seems to do just as good of a job. It calculates the delta between the new and old timeline and makes sure there aren't more changes in one request than can be changed in the interface. If it does, then the API 500s and the user gets a red flash error in the interface.

### DIFF
--- a/capstone/labs/views.py
+++ b/capstone/labs/views.py
@@ -1,4 +1,5 @@
 import json
+from dictdiffer import diff as dictdiff
 
 from django.shortcuts import render
 from django.http import JsonResponse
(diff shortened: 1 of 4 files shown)


capstone/labs/views.py
```diff
@@ -121,34 +122,49 @@ def chronolawgic_api_update(request, timeline_uuid):
...
+    ))
+
+    number_of_items_removed = len([thing[0] for thing in dictdiff(timeline_record.timeline, incoming_timeline)
+                             if thing[0] == 'remove'])
+
+    if number_of_items_modified > 1 or number_of_items_removed > 1:
+        return JsonResponse({'status': 'err', 'reason': "Timeline out of sync. More than one change or remove detected—"
+                                                        " aborted change protect timeline. Please refresh "
+                                                        "Chronolawgic. Did we mention this app is beta?😬"}, status=500)
+
```
anastasia, line 159: 😬 😀

anastasia (approved):

PR #1803, opened 2021-04-16
Cite extraction tweaks

jcushman:
* Include domain in all cite links
* Don't match roman numerals as page numbers


bensteinberg:
Wouldn't this still put the wrong domain in the cites if we're in beta mode?

jcushman:
Yes, but let's solve that later.

PR #1832, opened 2021-05-07
Update a JS package

bensteinberg:
I thought this would only make a change to `yarn.lock`, but `docker-compose exec web yarn upgrade url-parse` didn't make any change at all -- I had to run `docker-compose exec web yarn upgrade url-parse@^1.5.0` to make a new lock file, and it added a line to `package.json` as well. Let me know if there's a better way to do this.

### DIFF
--- a/capstone/package.json
+++ b/capstone/package.json
@@ -78,6 +78,9 @@
   "resolutions": {
     "vue-svg-loader/svg-to-vue": "0.6.0",
     "**/is-svg": "^4.2.2",
-    "**/ssri": "^8.0.1"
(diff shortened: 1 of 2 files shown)


jcushman:
See my comment on https://github.com/harvard-lil/capstone/pull/1754 -- I think you have to add the package you want to `resolutions`. (I can't remember if I tried adding it, running `yarn`, then removing it again.)

bensteinberg:
Thanks/sorry, I forgot about that. Let me give it a try.

jcushman:
Thanks!

PR #1833, opened 2021-05-12
Update postcss

bensteinberg:
https://github.com/postcss/postcss/blob/main/CHANGELOG.md


bensteinberg:
This may not be completely cooked:

https://evilmartians.com/chronicles/postcss-8-plugin-migration

I'm looking into it.

PR #1864, opened 2021-05-26
Add author

anastasia:
Also update snippets. @bensteinberg, this PR will require running `fab update_labs_snippets` that will create a new court snippet. It might also need back tracking/running migrations again if we run into the same problem (migrations having run, data not updated, like yesterday).


anastasia:
Screen Shot 2021-05-26 at 3 33 38 PM

ChefAndy (approved):
👍 Looks good

bensteinberg (approved):
+1

PR #1868, opened 2021-05-27
Search vuex

ChefAndy:
Done:
- Removed endpoint routing. Should be backwards-compatible with prior case search urls but not others, obviously.
- Added a vuex instance 
- Moved all data/logic that isn't strictly for single-component display into the vuex store. Removed almost all props.
- added vue-select without searching and multi-select because the styling would be too fussy
- slightly bodged styling where necessary

Left to do:
- implement vue-select features like multi-select and search for option
- improve accessibility for dropdowns— right now if something is selected, you can only tab to delete the selection and then you have to select something else rather than being able to just change it.
- vue-select components don't work with the query explainer


ChefAndy:
# tweaks
static/css/scss/labs-chronolawgic/base.scss
* z-index for error ribbon bumped

static/js/search/field-item.vue
* turned off input field highlighting when the input field was being used
* added clearable prop and option for vue-select

static/js/search/result-list.vue
* fixed janky formatting, fixed typo where > should have been >=

static/js/search/search-form.vue
* typo missing v-else
* added clearable prop to field-item

static/js/search/store.js
* changed download_size to page_size
* made a copy of values as they are at search to display search params

PR #1892, opened 2021-06-09
[Chronolawgic] quick little url validator

ChefAndy:


anastasia:
Cool! I think we should refactor sometime soon to make shared methods, but merging right now because this issue is definitely not specific only to this code. (also turns out I can update YOUR description in the PR? Wtf)

PR #1895, opened 2021-06-14
Search style/functionality

ChefAndy:
Various bug fixes and updates. This includes the multi-select drop downs for search filtering.

Simple pagination fix that was included might have been too simple— tweaking and submitting in a separate PR.


capstone/static/js/search/temp_court_list.js
```diff
@@ -0,0 +1,2 @@
...t."], ["va-high-ct-ch", "Va.: Va. High Ct. Ch."], ["va-spec-ct-app-1", "Va.: Va. Spec. Ct. App."], ["va-super-ct-ch", "Va.: Va. Super. Ct. Ch."], ["virginia-beach-cir-ct", "Va.: Virginia Beach Cir. Ct."], ["warren-cty-cir-ct-2", "Va.: Warren Cty. Cir. Ct."], ["washington-cty-cir-ct", "Va.: Washington Cty. Cir. Ct."], ["waynesboro-cir-ct", "Va.: Waynesboro Cir. Ct."], ["westmoreland-cty-cir-ct", "Va.: Westmoreland Cty. Cir. Ct."], ["william-cty-cir-ct", "Va.: William Cty. Cir. Ct."], ["williamsburg-james-city-cty-cir-ct", "Va.: Williamsburg & James City Cty. Cir. Ct."], ["williamsburg-cir-ct", "Va.: Williamsburg Cir. Ct."], ["winchester-cir-ct", "Va.: Winchester Cir. Ct."], ["winchester-corp-ct", "Va.: Winchester Corp. Ct."], ["wise-cty-cir-ct", "Va.: Wise Cty. Cir. Ct."], ["wythe-cty-cir-ct", "Va.: Wythe Cty. Cir. Ct."], ["vt", "Vt.: Vt."], ["vt-dist-ct", "Vt.: Vt. Dist. Ct."], ["w-va", "W. Va.: W. Va."], ["w-va-ct-cl", "W. Va.: W. Va. Ct. Cl."], ["wash", "Wash.: Wash."], ["wash-ct-app", "Wash.: Wash. Ct. App."], ["wash-dist-ct", "Wash.: Wash. Dist. Ct."], ["wis-ct-app-1", "Wash.: Wis. Ct. App."], ["wis", "Wis.: Wis."], ["wis-ct-app", "Wis.: Wis. Ct. App."], ["wyo", "Wyo.: Wyo."]];
```
jcushman, line 1: Was this meant to be replaced by something else? Or is the goal to temporarily include a hardcoded courts list and replace it later?

ChefAndy: I was going to have this here, now, just to keep things moving and replace it later with a snippet. This was intended to be a quicker PR than it ended up being, though, so maybe I should just add the snippet functionality now?

jcushman: Yeah at a minimum with this much of a shortcut I would include a comment about what it's a placeholder for.

If adding a snippet is a lift to do now, I'd also note that the [Django cache framework](https://docs.djangoproject.com/en/3.2/topics/cache/) might be just as good or better for some of what we use snippets for (where rendering the real thing only takes a second or two anyway) -- either as a `{% cache None "court_list" %}` tag in the template, or using `cache.get` in the view.

ChefAndy: Good call— my bad. 

It looks like template fragment caching plus helper functions could be a drop-in replacement for snippets as we use them— I'm looking at the details RN. I need to look for specifics, but there might be one or two snippets for which that first load might be kind of a doozy. I think we could get around that by not automatically expiring caches and maybe having a fab task that runs immediately after deploy that requests pages with heavier data processing.

jcushman: Maybe we use cache for things that run in under a second like "dump 1000 entries for a dropdown list", and write files to /download/ for statistical reports that take longer to run. We didn't have the download section when we made the snippets thing, but now that we do it makes sense to write statistics like cases_by_decision_date to a csv in a statistics folder in downloads. Then the link can just be to that rather than to a database view. Let's chat about this offline as far as architecture ...

PR #1905, opened 2021-06-17
Update Reverse Sort, Replace Snippets with Cached Template Tag

ChefAndy:
Two things I discussed with @jcushman in the last PRs:

I fixed the prev link bug and I simplified the reverse sorting function in the API a bit.

Also, I replaced snippets with the built-in django caching. Initially I was going to replace them with template fragment caching, but they seem to only currently be used for pre-rendering JSON objects which didn't make sense to render in a template, so I replaced them with a template tag that internally uses Django's built-in cache. Maybe this caching would, at some point, be better handled in the view?  I think this is a fine solution for now, at least. 

In a template, the usage is exactly the same, but instead of the `{% snippet 'map_numbers' '{}' %};` template tag use `{% cached_json_object 'map_numbers' '{}' %};`. 

@bensteinberg I removed the Model, made a migration, removed the fab tasks and the celery task that called them. Seems pretty straightforward but I rarely touch that stuff so I wanted another set of eyes.

@anastasia as someone who uses snippets every now and then I'm happy to hear any input you've got.


ChefAndy:
Going to write a couple more tests

capstone/capweb/templatetags/cached_json_object.py
```diff
@@ -0,0 +1,80 @@
...
+            for jurisdiction in Jurisdiction.objects.order_by('slug').all()
+            if jurisdiction.slug != 'regional']
+    return json.dumps(jurisdictions)
+
+def search_court_list():
+    output = []
+    for court in Court.objects.order_by('slug').all():
+        try:
+            output.append((court.slug, "{}: {}".format(court.jurisdiction, court.name)))
+        except Jurisdiction.DoesNotExist:
```
jcushman, line 68: Courts without a jurisdiction are an error, so I think we want to avoid having any of those rather than checking for them here. I just fixed the three that somehow didn't have one.

This query also wants a `select_related('jurisdiction')` to avoid doing a jurisdiction query for each court.

capstone/capweb/templatetags/cached_json_object.py
```diff
@@ -0,0 +1,80 @@
...
+    """)
+    # get column names from sql query
+    cols = [col[0] for col in cursor.description]
+    # create output where each key is a jurisdiction and each value is a dict of values from the sql query
+    output = {row[0]: dict(zip(cols[1:], row[1:])) for row in cursor.fetchall()}
+    return json.dumps(output)
+
+
+def search_jurisdiction_list():
+    jurisdictions = [ (jurisdiction.slug, jurisdiction.name_long)
```
jcushman, line 58: This can be `jurisdictions = Jurisdiction.objects.order_by('slug').exclude(slug='regional').values_list('slug', 'name_long')`

capstone/capweb/templatetags/cached_json_object.py
```diff
@@ -0,0 +1,80 @@
...
+          COUNT(c.id) AS case_count,
+          COUNT(DISTINCT c.volume_id) AS volume_count,
+          COUNT(DISTINCT c.reporter_id) AS reporter_count,
+          SUM(CASE WHEN (c.first_page||c.last_page)~E'^\\d+$' THEN c.last_page::integer-c.first_page::integer+1 ELSE 1 END) AS page_count
+        FROM capdb_jurisdiction j 
+          LEFT JOIN capdb_casemetadata c ON j.id=c.jurisdiction_id 
+        WHERE
+          c.in_scope IS True
+        GROUP BY j.id;
+    """)
```
jcushman, line 49: This takes 100 seconds to run, so it doesn't work to do it in a view. I think this is like the tsv files -- we want to add it to a `/download/statistics/latest/jurisdiction_counts.json` file and load it from there.

capstone/scripts/update_snippets.py
```diff
@@ -1,181 +0,0 @@
...
-from capweb.templatetags.api_url import api_url
-from tqdm import tqdm
-
-
-def update_all():
-    update_map_numbers()
-    cases_by_jurisdiction_tsv()
-    cases_by_reporter_tsv()
-    cases_by_decision_date_tsv()
-    search_reporter_list()
```
jcushman, line 18: These tsv snippets were no longer used, huh, in favor of static files last updated three years ago? 😬 I think we want to keep the code for these but have a task to write them to `/download/statistics/`, probably at the same time we do a bulk data export.

jcushman (changes requested):
Thanks for working on this. The snippets part still needs some work to be performant here (see comments), including figuring out how to write statistics files to the downloads section, so could you maybe split the API fix out into a separate PR? Like a `git merge --no-commit --no-ff` of this branch onto the latest develop, and then send PRs for the two blocks of work separately.

---

You're right that doing the caching in the view would remove some moving parts. I think you would just do something like this:

```
# capweb.helpers:

@cache_func(lambda: "search_jurisdiction_list", settings.CACHED_JSON_OBJECT_TIMEOUT)
def search_jurisdiction_list():
    ...

# view:
    ...
    'search_jurisdiction_list': search_jurisdiction_list(),
```

Strikes me as a little cleaner. 🤷‍♂️

---

The big thing here is we can't do long-running queries like the one in `map_numbers` this way, and I don't think we want to delete the snippets for the statistics tsvs and be left serving the static ones generated in 2018. Let's fix both of those by figuring out a clean way to dump statistics in the /download/ section of the site ... come to think of it, maybe the clean way is to add a final step to `scripts.export.init_export` that creates a statistics folder as a subfolder of a bulk export? This might also be a good opportunity to rethink what wants to go in that folder as well.

capstone/capweb/templatetags/cached_json_object.py
ChefAndy: > This takes 100 seconds to run, so it doesn't work to do it in a view. I think this is like the tsv files -- we want to add it to a `/download/statistics/latest/jurisdiction_counts.json` file and load it from there.

Right— Ah, ok. I need to get shell access squared away so I can profile some of this stuff. My initial plan to mitigate this problem— which I didn't end up implementing— was having a function in our prod deployment routine that would hit any page using a particularly long-running query, which should populate Redis, right? I don't think we'd need to automatically expire these things because they shouldn't need to be updated any more frequently than snippets. If Redis ended up losing its state for some reason, it would at least be a self-fixing problem even if not immediately (though we should probably log instances of cached values being regenerated so we could monitor it.) Does that make sense or should I just scrap it?

capstone/scripts/update_snippets.py
ChefAndy: > Thanks for working on this. The snippets part still needs some work to be performant here (see comments), including figuring out how to write statistics files to the downloads section, so could you maybe split the API fix out into a separate PR? Like a `git merge --no-commit --no-ff` of this branch onto the latest develop, and then send PRs for the two blocks of work separately.
> 
> You're right that doing the caching in the view would remove some moving parts. I think you would just do something like this:
> 
> ```
> # capweb.helpers:
> 
> @cache_func(lambda: "search_jurisdiction_list", settings.CACHED_JSON_OBJECT_TIMEOUT)
> def search_jurisdiction_list():
>     ...
> 
> # view:
>     ...
>     'search_jurisdiction_list': search_jurisdiction_list(),
> ```
> 
> Strikes me as a little cleaner. 🤷‍♂️
> 
> The big thing here is we can't do long-running queries like the one in `map_numbers` this way, and I don't think we want to delete the snippets for the statistics tsvs and be left serving the static ones generated in 2018. Let's fix both of those by figuring out a clean way to dump statistics in the /download/ section of the site ... come to think of it, maybe the clean way is to add a final step to `scripts.export.init_export` that creates a statistics folder as a subfolder of a bulk export? This might also be a good opportunity to rethink what wants to go in that folder as well.

Man— I scoured the code for current uses of these snippets and totally missed that. I haven't even thought about these things since I wrote the initial script in like 2017. Damn. I assumed it was rolled up into our other data export functionality. I'll restore this processing code and have a task write them out to files. Easy fix.

capstone/capweb/templatetags/cached_json_object.py
ChefAndy: Good call.

capstone/capweb/templatetags/cached_json_object.py
jcushman: Yeah I think that's going to be jankier since it has a bunch of assumptions and moving parts built in -- assumes that redis isn't cleared frequently which is otherwise harmless, assumes that the caching step will complete even when the downstream request is killed after 1 minute, implies at least a couple minutes of error messages during regeneration (which in turn have to be checked to make sure they're not real errors), etc. Writing the file out to /download/statistics/ avoids any jank, and also comes with the upside that the file is potentially useful to people.

ChefAndy: > Writing the file out to /download/statistics/ avoids any jank, and also comes with the upside that the file is potentially useful to people.

Easy enough!

PR #1917, opened 2021-06-24
quick script to pull dupes from a body of simhashes

4ndygu:
I used this script to pull duplicates out of a subset of all cases, but am hoping to parallelize or at least run on a hosted instance later this summer. Leverages http://benwhitmore.altervista.org/simhash-and-solving-the-hamming-distance-problem-explained/?doing_wp_cron=1624539467.8441119194030761718750 to get simhash comparison in what seems to be O(nlogn)


jcushman:
Sorry, I know I suggested sending a PR for this but it looks like a hassle to get the tests passing. Maybe simplest to comment out the file contents to avoid having to satisfy flake8 and pytest? Otherwise we have to either add this to a pytest exclusion list or add the simhash dependency, and those both seem more invasive than necessary until we pick this script back up to work on. 

Could also create a new issue for duplicate detection and paste this code in there instead of adding it to scripts.

4ndygu:
sure! i fixed the linter errors and renamed the file to txt.

jcushman:
Cool, thanks!

PR #1918, opened 2021-06-24
Prev fix

ChefAndy:
Tests are erroring out locally— I think it's a local problem. Putting in a draft PR to make sure.

### DIFF
--- a/capstone/capapi/pagination.py
+++ b/capstone/capapi/pagination.py
@@ -15,7 +15,6 @@ from rest_framework import serializers
 
 from capapi.resources import CachedCountQuerySet
 
-
 class CapPagination(CursorPagination):
     # This should be larger than the max number of records that share the same ordering field, such as decision_date,
     # but not too much larger to avoid allowing needlessly expensive queries.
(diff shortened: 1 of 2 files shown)


ChefAndy:
The errors are, as suspected, local. (had to kill the hyperscan cache— new one for me.)

jcushman:
Thanks for taking a pass at this. I think what you had here also wasn't quite right -- it was reversing `queryset._sort` from `[{'_score': {'order': 'desc'}}, {'decision_date_original': {'order': 'asc'}}, 'id']` to `[{'_score': {'order': 'asc'}}, {'id': {'order': 'desc'}}, {'decision_date_original': {'order': 'desc'}}, {'id': {'order': 'desc'}}]`, adding an extra `id` after each sort field instead of reversing the existing one, which caused an error when trying to reverse the default sort on `/v1/cases/`.

Given how tricky this bit of code turns out to be, I decided to break it out into its own helper function with doctests, and also update the pagination test to lock down how it behaves including when there's no sort order provided.

PR #1925, opened 2021-06-29
quick fix creating events/cases on new timelines, removing cases

anastasia:


capstone/labs/models.py
```diff
@@ -71,7 +71,7 @@ def first_year(case_years, event_years):
         elif case_years:
             return min(case_years)
         elif event_years:
-            return min(case_years)
+            return min(event_years)
```
ChefAndy, line 74: 😑

PR #1930, opened 2021-07-09
Add support for citation trendlines

4ndygu:
This PR adds support for querying the `/trends/` API by citation graphs over time, like the following:

Screen Shot 2021-07-09 at 12 43 43 PM

Screen Shot 2021-07-09 at 12 50 20 PM

In order to trigger this functionality, a must must append the `@` symbol to the end of their case ID. There is current support for all *specific* jurisdictions, as well as lookups. However, support for wildcard jurisdictions for cases has been removed to minimize load.

Screen Shot 2021-07-09 at 12 50 58 PM


4ndygu:
after some great advice from Jack, `*` is now supported, and the backend is rewired from the API to Elasticsearch directly.

4ndygu:
This PR now adds a `facet` feature to the API, which permits aggregation natively in the API. The /trends/ endpoint also now supoprts a limited domain specific language (DSL) to directly filter for specific items in the `/trends/` endpoint. 

Screen Shot 2021-07-15 at 12 03 25 PM

TODO:
Support gt, lt, range, and `in` operators for bulk upload.
Build a converter from search -> trends
Build a converter from the case view -> trends.
tests

PR #1932, opened 2021-07-14
Snippet Removal Part 1— TSV generation to download_storages

ChefAndy:
This part just moves the TSV snippets to regular files. Part 2 with the caching coming up.


jcushman (changes requested):
Nice start! I think rather than storing these in the base downloads directory, we should store them with each bulk export. You could update `scripts.export.init_export` to call `scripts.update_download_tsv.update_all` at the end, with a path of `output_path / 'statistics'`. And then the about.md links can go to `{% url "download-files" "bulk_exports/latest/statistics/..." %}`

PR #1933, opened 2021-07-19
api(), allow click in, set nan -> 0

4ndygu:
Allow empty `api()` calls
allow users to click into the timeines
set NaN -> 0


4ndygu:
@jcushman

PR #1937, opened 2021-07-21
Support for author and author disposition fields

4ndygu:
Note: This PR introduces two schema changes to the CAP dataset:

- casebody_data.text.opinions.author is now an FTSField.
- casebody_data.text.opinions is now a nested field to support querying a opinions with multiple attributes, such as an opinion with judge X && judge X dissenting.

This PR introduces two additional fields for filtering against the CAP dataset: `author` and `author_disposition`. Author is a string field that merely scans for tokens of a particular justice:

Screen Shot 2021-07-21 at 10 36 44 AM

`author_disposition` is a colon-separated field which can filter for a justice and their particular disposition on a case. For instance, to look for all Scalia dissents:

Screen Shot 2021-07-21 at 10 37 15 AM


4ndygu:
still need to lint and fix an issue with search text highlights, but --
- this query now uses a new backend, NestedSimpleStringQueryBackend, that uses a nested query to make separate simple string queries against opinions cases. MultiFieldFTS is now changed as well, as otherwise object level filtering per the old behavior would have failed.
- trends are also supported for these fields.

todo:
- lint / test
- add UI support for text fields

PR #1938, opened 2021-07-22
don't add inner_hits on a second nested query

4ndygu:
this allows multiple author fields to construct a logical AND relationship. the highlight inner_fields stanza would only appear attached to the first nested query, but this should not change the highlighting of results under the hood.


4ndygu:
never mind -- it is both simpler and just as good to simply have random names on each inner_hits field.

PR #1939, opened 2021-07-23
support cites_to_X

4ndygu:
To answer questions such as citations to a particular class of queries (filtered by jurisdiction, court, etc), this PR supports extraction of a class of queries via a first query (sped up with source=false and async connection pools), then uses the IDs in a second level terms query.

The max number if matched citations is currently 20K for stability purposes.

example:
http://api.case.test:8000/v1/cases/?cites_to_jurisdiction=us&cites_to_court=us&jurisdiction=cal


4ndygu:
Additional features since PR open:
- support for nested fields (cites_to_author, cites_to__author_disposition)
- random sampling of caseid extraction to hopefully capture caseids evenly when the original set > 20K cases.
- manual edit of hashes in requirements.txt to deal with strange pip redundancy fixes ruining the hash feature

capstone/capapi/resources.py
```diff
@@ -159,6 +162,57 @@ def apply_replacements(item, replacements, prefix="[ ", suffix=" ]"):
...
+
+    async def get_query_results(query_body):
+        es = AsyncElasticsearch([settings.ELASTICSEARCH_DSL['default']['hosts']])
+
+        await asyncio.gather(*[
+            asyncio.ensure_future(fetch(es, i, query_body))
+            for i in range(0, max_workers)
+        ])
+
+        es.close()
```
4ndygu, line 207: todo: AWAIT es.close()

PR #1941, opened 2021-07-27
Handle socket.gaierror in is_google_bot()

bensteinberg:


bensteinberg:
This is to reduce noise during what I believe are transient DNS problems. They take the form of `socket.gaierror: [Errno -2] Name or service not known` or `socket.gaierror: [Errno -3] Temporary failure in name resolution`.

bensteinberg:
It's possible that the DNS problems are due to rate-limiting, though that wouldn't explain the episodic quality of this error, given our traffic.

PR #1954, opened 2021-07-29
Use per process searches for parallel_search

4ndygu:


capstone/capapi/resources.py
```diff
@@ -186,47 +186,60 @@ def remove_nested_keys(target_dict, keys):
...
+def run_search(search_blob):
+    # get elasticsearch query as a dictionary
+    if search_blob['remove_keys']:
+        remove_nested_keys(search_blob['query_body'], search_blob['remove_keys'])
+
+    es = Elasticsearch(search_blob['host'])
+    resp = es.search(index=search_blob['index'], body=search_blob['query_body'])
+    hits = deep_get(resp, ['hits', 'hits'])
+    es.close()
+    return search_blob['worker_i'], [hit['_id'] for hit in hits]
```
4ndygu, line 198: this makes some assumptions about what exactly we want from the search, but that limitation may be fine for now.

PR #1962, opened 2021-08-03
Support queries against citations for a particular *opinion*

4ndygu:
just for visibility / review, tests and polish needed due to the large structural change.

This PR allows users to query against a particular *opinion* when supplying either the `author` or `author_type` parameters in conjunction with the `cites_to`, `cites_to_id`, `cites_to.reporter`, or `cites_to.category` parameters (prefixed with opinion__) against the V1 Cases API. 

To implement this, this PR makes the following changes:
- when `fab refresh_case_body_cache` is called, `extract_citations` is changed such that the relevant index of the opinion for a particular citation is retrieve and deposited into `capdb_extractedcitation`. 
- when loading postgres into elasticsearch, extracted citations are now a **nested** field inside `casebody_data.text.opinions`. 
- [ not as confident ] -- within each resolved cluster generated by eyecite in `extract_citations`, create a citation for each `opinion_id`.

This requires us to use nested queries in cites_to, cites_to_id, cites_to.category, and cites_to.reporter, as well as multi-level nested queries for querying citations *within* a particular opinion. Make sure to benchmark the additional footprint.


capstone/capapi/views/api_views.py
```diff
@@ -236,9 +254,11 @@ def filter_queryset(self, queryset):
...
                 data_formats_to_exclude.remove('text')
 
-            source_filter = {"excludes": ["casebody_data.%s" % format for format in data_formats_to_exclude]}
-        else:
-            source_filter = {"excludes": "casebody_data.*"}
+        source_filter = {
+            "excludes": ["casebody_data.%s" % format for format in data_formats_to_exclude],
+        }
+
+        # raise Exception(source_filter)
```
4ndygu: kill this later

capstone/capapi/filters.py
```diff
@@ -423,7 +444,21 @@ class NestedFTSFilter(BaseSearchFilterBackend):
...
+    combine_with_cite = False
     matching = MATCHING_OPTION_MUST
+    search_key = 'query_values'
+
+    def get_search_query_params(self, request, view):
+        params = super().get_search_query_params(request)
+
+        if self.combine_with_cite and params:
+            # Add params from multinestedfilteringbackend. View not needed.
+            cite_params = MultiNestedFilteringFilterBackend().get_filter_query_params(request, view, called_from_multi=True)
```
4ndygu: manipulate inbound requests first and read a `opinion_` prefix to each relevant feature

capstone/scripts/extract_cites.py
```diff
@@ -45,29 +45,31 @@ def extract_citations(case, html, xml):
...
-            normalized_forms['rdb_normalized_cite'] = normalize_cite(normalized_forms['rdb_cite'])
-            eyecite_cite.normalized_forms = normalized_forms
-
-            for k, v in normalized_forms.items():
-                cites_by_type[k].append(v)
-
-        if extracted_cites:
-            extracted_els.append((el.attr("id"), el_text, extracted_cites))
+    # Extract cites from each paragraph for each opinion:
+    for i, section in enumerate(clean_html_pq('.opinion').items()):
```
4ndygu: assume that we don't care about citations that aren't made from opinions. will correct this otherwise.

capstone/capdb/tests/test_tasks.py
```diff
@@ -328,14 +328,17 @@ def test_extract_citations(reset_sequences, case_factory, elasticsearch, client)
 
     # check API response
     case_json = client.get(api_reverse("cases-detail", args=[case.id])).json()
+
+    print("CASE JSON IS")
+    print(case_json['cites_to'])
     assert case_json['cites_to'] == [
+        {'cite': '2 US 2', 'category': 'reporters:federal', 'reporter': 'U.S.', 'case_ids': [2, 3], 'weight': 2},
```
4ndygu: due to the way we collate the citations post search, this order is reversed.

capstone/capdb/models.py
```diff
@@ -2055,6 +2072,7 @@ class ExtractedCitation(models.Model):
     category = models.CharField(max_length=255, blank=True, null=True, help_text="Source and cite_type from reporters-db, e.g. 'laws:leg_statute'")
     weight = models.SmallIntegerField(default=1, help_text="Number of citations in cited_by, counting short cites.")
     year = models.SmallIntegerField(null=True, help_text="Year included in citation.")
+    opinion_id = models.SmallIntegerField(null=True, blank=True, help_text="Id of opinion making extracted citation.")
```
4ndygu, line 2075: assume there will be never be 32767 opinions lol

capstone/capdb/models.py
```diff
@@ -1476,6 +1477,22 @@ def redact_obj(self, text):
...
         return apply_replacements(text, replacements)
 
+    def insert_citations(self, text):
+        citations = json.loads(serializers.serialize('json', ExtractedCitation.objects.filter(cited_by_id=self.id)))
+
+        for citation in citations:
+            opinion_id = citation['fields']['opinion_id']
+            fields = citation['fields']
+            # TODO: Consider a better solution to manual casting
+            fields['target_cases'] = [int(id) for id in json.loads(fields['target_cases'])]
```
4ndygu: currently, we use a Django serializer to convert the extractedcitation back into a JSON object that can enter the elasticsearch cluster. i'm curious about whether there is another method to cast that is less high touch

4ndygu:
should be ready for a review. Examples:

http://api.case.test:8000/v1/cases/?opinion__cites_to_id=1785580&author_type=newby:concurrence&name_abbreviation=state%20v.%20defoe

This ^ above query shows results because Justice Newby's concurrence is the only opinion that cites Mapp v. Ohio. If we try something like:

http://api.case.test:8000/v1/cases/?opinion__cites_to_id=1785580&author_type=timmons-goodson:majority&name_abbreviation=state%20v.%20defoe

or

http://api.case.test:8000/v1/cases/?opinion__cites_to_id=1785580&author=timmons-goodson&name_abbreviation=state%20v.%20defoe

the apiserver would return no results. However, 

http://api.case.test:8000/v1/cases/?cites_to_id=1785580&author=timmons-goodson&name_abbreviation=state%20v.%20defoe

returns results because it references the entire **case**. name_abbreviation is only meant to show compatibility with existing queries.

capstone/capapi/serializers.py
```diff
@@ -150,7 +150,9 @@ def as_dict(obj):
 
         # get extracted_citations list, removing duplicate c["cite"] values
         extracted_citations = []
-        for c in s["extracted_citations"]:
+        ec = [o['extracted_citations'] for o in s['casebody_data']['text']['opinions'] if 'extracted_citations' in o]
+        ec = [item for sublist in ec for item in sublist]
+        for c in ec:
```
jcushman, line 155: Do we want to include `"opinion_index": <index>` in the extracted_cite output?

capstone/capapi/filters.py
```diff
@@ -357,14 +358,22 @@ def construct_search(cls, request, view, search_backend):
...
 
         query_operator = operator.or_
         if search_backend.matching == MATCHING_OPTION_MUST:
             query_operator = operator.and_
 
-        for search_term in query_params:
+        search_terms = query_params
+        if type(query_params) == dict:
+            search_terms = query_params[search_backend.search_key]
+            del query_params[search_backend.search_key]
```
jcushman: can be `search_terms = query_params.pop(search_backend.search_key)`

capstone/capapi/views/api_views.py
```diff
@@ -227,6 +243,9 @@ def get_serializer_class(self, *args, **kwargs):
     def filter_queryset(self, queryset):
         queryset = super(CaseDocumentViewSet, self).filter_queryset(queryset)
         # exclude all values from casebody_data that we don't need to complete the request
+        data_formats_to_exclude = ["text.head_matter", "text.corrections", "text.parties", "text.attorneys", 
+            "text.opinions.text", "text.opinions.type", "text.opinions.author", "text.judges", "html", "xml"]
```
jcushman: Can this be done with something like `"excludes": ["casebody_data.text"], "includes": ["casebody_data.text.opinions.extracted_citations"]`? Listing all the fields except one feels brittle.

capstone/capdb/models.py
jcushman: Haven't tested, but something like this should work 

```
    def prepare_casebody_data(self, instance):
        body = instance.body_cache
        text = deepcopy(body.json)
        serializer = self._fields['casebody_data']['text']['opinions']['extracted_citations']
        cites_by_id = group_by(instance.extracted_citations.all(), lambda c: c.opinion_id)
        for i, opinion in text['opinions']:
            if i in cites_by_id:
                opinion['extracted_citations'] = [serializer.get_value_from_instance(c) for c in cites_by_id[i]]
        return instance.redact_obj({
            'xml': body.xml,
            'html': body.html,
            'text': text,
        })
```

Note by the way that `ExtractedCitation.objects.filter(cited_by_id=self.id)` wouldn't be performant here -- indexing a volume in elasticsearch should take `O(1)` sql queries, which is handled by a `prefetch_related('extracted_citations')` hiding in `update_elasticsearch_for_vol`. That means we get `instance.extracted_citations.all()` for free, but any other queries to get the same data would hit the DB.

capstone/scripts/extract_cites.py
jcushman: Unfortunately I think we do -- if a case has a lot of head matter you'd like to be able to click from citations in the head matter to read the cases. I'm not sure where you store those in the index though.

capstone/scripts/extract_cites.py
```diff
@@ -209,24 +217,31 @@ def resolve_full_cite(eyecite_cite):
...
+                **normalized_forms,
+                reporter=reporter,
+                category=category,
+                cited_by=case,
+                target_case_id=target_case_id,
+                target_cases=target_cases,
+                groups=groups,
+                metadata=metadata,
+                pin_cites=pin_cites,
+                weight=weight,
```
jcushman, line 235: pin_cites and weight are wrong here -- they'll redundantly include all the pin_cites for all opinions.

This also makes me think it would be tricky with the nested structure to frame a query like "get cases that cite this case ordered by how many times they cited to it" ... maybe that gets denormalized somewhere.

capstone/capweb/templates/docs/02_site_features/03_api.md
```diff
@@ -133,13 +133,22 @@ Many parameters can be appended with `__in`, `__gt`, `__gte`, `__lt`, or `__lte`
...
-    * __description:__  find cases that cite to the given citation or case ID, e.g. `1 Ill. 21` or `12345`
+    * __description:__  find cases that cite to the given citation or case ID, e.g. `1 Ill. 21` or `12345`. 
 * `cites_to.reporter`
     * __data type:__    string
     * __description:__  find cases that cite to the given reporter, e.g. `Ill. 2d` or `Harv. L. Rev.`
 * `cites_to.category`
     * __data type:__    string
-    * __description:__  find cases that cite to the given citation category, e.g. `reporter:state` or `reporters:federal`. Category values can be found in results for the cases endpoint under `extracted_citations[].category`
+    * __description:__  find cases that cite to the given citation category, e.g. `reporter:state` or `reporters:federal`. Category values can be found in results for the cases endpoint under `extracted_citations[].category`.
+* `opinion__cites_to`
```
jcushman: We should chat offline about whether to add this prefix or just say that `author` is ANDed with `cites_to` (and `search`?). I don't quite have it figured out.

jcushman (changes requested):
Cool! I have barely enough brain for this at the moment but I threw in some initial comments.

capstone/capapi/views/api_views.py
4ndygu: i agree that its not optimal to hard code these fields. i did try that exact solution, but the exclusion filter takes precedent over the inclusion filter, rendering the above result into a null.

capstone/capapi/serializers.py
4ndygu: that info seems encoded in the blob anyways. will remove.

capstone/capapi/filters.py
4ndygu: eagle eye

capstone/capweb/templates/docs/02_site_features/03_api.md
4ndygu: yeah, that conversation makes sense - i originally had that logic here, but was scared to miss the scenario where users are only curious about citations to a case as some signal for a topic that they want to filter on and then retrieve the opinions of a particular justice, versus the scenario where we want to think about when a justice *makes a cite*. happy to learn a bit more about the problem space

capstone/capdb/models.py
4ndygu: thanks! i'm snagging something like this to also filter on the `cited_by_id` without eating a second query against the db:
```
    def prepare_casebody_data(self, instance):
        body = instance.body_cache
        serializer = self._fields['casebody_data']['text']['opinions']['extracted_citations']
        # We must unroll to dict to support the `in` operator
        outbound_cites = [c for c in ExtractedCitation.objects.all() if c.cited_by_id == instance.id]
        cites_by_id = {k: list(v) for k,v in groupby(outbound_cites, lambda c: c.opinion_id)}

        for i, opinion in enumerate(body.json['opinions']):
            if i in cites_by_id:
                body.json['opinions'][i]['extracted_citations'] = [serializer.get_value_from_instance(c) for c in cites_by_id[i]]

        return instance.redact_obj({
            'xml': body.xml,
            'html': body.html,
            'text': body.json,
        })
```

capstone/capapi/serializers.py
4ndygu: this now pops the opinion_id.

capstone/capapi/filters.py
4ndygu: done

capstone/scripts/extract_cites.py
4ndygu: i understand the issue with pin_cites -- how do you feel about only storing the `pin_cites` value for the first entry? I believe this should still support the search you're thinking about. To separate out the problem, we're looking for a query like this:

```
nested query A -> bool/must -> another nested query B with filters for extracted_citations.field = somevalue
```

were we to extend the function, i would imagine us doing:
1. index another field on the first `extracted_citation` value in the cluster that represents the number of documents in the nested structure (we can use script-based sorting, but its not as efficient.)
2. add another query within the **outer** nested structure, kinda like so:

```
nested query A -> bool/must -> another nested query B with filters for extracted_citations.field = somevalue
                                             \_-> another nested query C that checks where extracted_citations.normalized_cite = {our_identifier} and extracted_citations.citecount > {our_threshold}
```

there is certainly a little more complexity here, though I'm not sure what % is designed by us and what % is inherent to how elasticsearch has approached nesting in general.

can you dig into the issue with weight again? It looks to me like weight is just the size of the cluster, with the assumption being that more frequently cited items are more weighty. I suppose there is a limited denormalization here because they're duplicates, but couldn't we say the same about all the other fields? (reporter, category, cited_by, target_case_id, target_cases, groups, metadata)?

capstone/scripts/extract_cites.py
4ndygu: dangit!

I may be wrong here, but what if we keep the annotations that seem to be getting written at https://github.com/harvard-lil/capstone/blob/develop/capstone/scripts/extract_cites.py#L126 but refuse to add them to the metadata? It seems significant to be able to distinguish when the head matter is making the citation v. when the judge is making the citation. assuming that the above code captures all the usefulness of our processing of citations in head matter, it might make sense to annotate but not count it as an extracted citation.

Otherwise, maybe we can place the citation where it was previously and rename the key `other_citations` / `head_citations` or something else. I'm not sure if we should still support queries against those citations or count them as part of the weight. In my mind they certainly seem less weightier, but this would exclude someone doing metaanalysis of head matter, i suppose.

4ndygu:
changes:
- exclude fields dynamically pulled from `._fields`
- opinion__ prefix -> author__
- instance.extract_citations.all() called.

todo:
- handle citations in head_matter
- more effective pin cites

capstone/scripts/extract_cites.py
4ndygu: changed this logic to:
- extract pin_cite at time of citation.
- extract weight after a `group_by` operation that splits each cluster by opinion_id.

4ndygu:
head_matter is now included in an opinion in elasticsearch, but rendered back into the normal `head_matter` section on render.

4ndygu:
new changes:
- quick bug fix, whereby groupby is only given sorted collections of objects
- Search UI is changed so when an `author` or `author_type` field is populated, the `opinion__cites_to` field fades in. Vanilla `cites_to` is also added.

capstone/capapi/serializers.py
```diff
@@ -167,6 +169,15 @@ def as_dict(obj):
                 extracted_cite['pin_cites'] = c['pin_cites']
             extracted_citations.append(extracted_cite)
 
+        # move head_matter outside of casebody_data
+        head_matter = list(filter(lambda x: x['type'] == 'head_matter', s['casebody_data']['text']['opinions']))
+        head_matter = head_matter[0] if head_matter else []
+        if head_matter:
+            s['casebody_data']['text']['opinions'].remove(head_matter)
+
+        if 'text' in head_matter:
```
jcushman, line 180: Is this `if 'text' in head_matter:` block right? I would think it would want to be `s['casebody_data']['text']['head_matter'] = head_matter['text']` within the `if head_matter:` block, but I could be misreading.

capstone/capapi/serializers.py
jcushman: I mean, do we want to add  `"opinion_index": <index>-1` to `extracted_cite` in the next lines? (`-1` to account for moving `head_matter` around.) Otherwise there can be multiple copies of the same cite in the API with no way to tell why.

capstone/scripts/extract_cites.py
```diff
@@ -195,38 +205,46 @@ def resolve_full_cite(eyecite_cite):
...
-            metadata=metadata,
-            pin_cites=pin_cites,
-            weight=weight,
-            year=first_cite.year,
-        )
-
-        found_cites.append(extracted_cite)
+        for opinion_id, opinion_cluster in opinion_clusters.items():
+            opinion_includes_cite = {}
+            for cite in opinion_cluster:
```
jcushman, line 211: Loop isn't necessary anymore, right? You just end up processing `opinion_cluster[0]` for each cluster.

jcushman (commented):
Almost there, just a few more Qs.

capstone/capapi/serializers.py
4ndygu: looool, documents.py now includes the opinion id.

4ndygu: src: https://github.com/harvard-lil/capstone/pull/1962/files#diff-464cbe1adc7d1bb94b3b7c4216797eef8f49312e8a58d5f6fd2a853e5842ef64R171

capstone/scripts/extract_cites.py
4ndygu: yeah -- we now only process opinion_cluster[0] to pull the weight, but use cluster[0] to pull the other metadata. we loop through the opinion_cluster for pin cites.

4ndygu:
this is definitely my fault, but may make sense to squash and merge this one if we can

PR #1968, opened 2021-08-09
Add UI Search elements

4ndygu:
This PR adds the following UI Search elements:

- cites_to
- author
- author_type
- author__cites_to (appears only if author or author_type are filled)
- cites_to__* (is a dropdown, which exposes a bottom field)
Screen Shot 2021-08-09 at 1 18 50 PM

Animation currently not enabled for cites_to__* to avoid strange nested objects in the DOM. this is resolvable with something like:
```
            <transition-group tag="div" name="fadegroup" class="search-field shown">
                <div v-for="dynamicField in $store.getters.getExposeDynamicCitesToField" :key=dynamicField.name class="search-field shown">
                  <field-item :clearable="true" :multiple="true" :field="dynamicField"></field-item>
                </div>
            </transition-group>
```

which is currently in progress.


4ndygu:
do not merge prior to https://github.com/harvard-lil/capstone/pull/1962

4ndygu:
@jcushman should be ready for review.

4ndygu:
actually, messing with some subtle ux bugs:

- adding new 'cites_to__' items does not seem to save the stored `value` of the item.

4ndygu:
now ready

PR #1980, opened 2021-08-23
Use `and` default_operator for simple search queries

4ndygu:
hardcoding because
 - this applies for all supported backends given how we use the simple search query filter
 - a small amount of confusion would be generated with the `matching` attribute as applied to _multiple_ filters that use this backend.


4ndygu:
this test should now work. the con is that the test fix hides the antipattern that search no longer works across multiple opinions. fix for that to be discussed elsewhere

PR #1999, opened 2021-09-14
Add scheme to context and command line

bensteinberg:


bensteinberg:
This is wrong:

Screen Shot 2021-09-14 at 4 52 17 PM

PR #2035, opened 2022-01-08
Bug Fixes

4ndygu:
Sorry about the delay, was fighting burnout + getting reacquainted with the code.

This PR fixes a few bugs:
- The screen now raises a custom error if a given query returns > 20K results (I don't have 20K cites on the test cite, so I simply changed the threshold to force the error for this test).

Screen Shot 2022-01-07 at 3 58 55 PM

- I changed the controllers' management of `ValidationError` to throw the `ValidationError` directly rather than centralize the error throw via return codes. This should be shorter and less complex, but lmk if you would like to change it back.

- There was a bug where click-ins into the timeline weren't working due to the misplacement of `urls` rather than `this.urls`. This PR fixes that as well.

### DIFF
--- a/capstone/capapi/filters.py
+++ b/capstone/capapi/filters.py
@@ -4,6 +4,7 @@ import six
 import re
 import uuid
 
+from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.functional import partition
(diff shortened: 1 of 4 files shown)


jcushman:
Sorry for the delay in reviewing! This looks great.

PR #2045, opened 2022-02-02
Update Elasticsearch and node

bensteinberg:
https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-7.17.0.html

https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md#14.19.0


bensteinberg:
@jcushman can you think of a reason why, in the export tests, a case_factory case would have changed from the expected "open" to "restricted"?

jcushman:
Hopefully fixed initial test failure in a separate PR ...

PR #2048, opened 2022-02-09
Updates for devs with non-amd64 machines, catch-up for move to GitHub Actions

bensteinberg:
So far, this just updates the capstone image -- changes for the postgres image will follow, with any luck. https://github.com/docker-library/postgres/issues/549 exposes what I think is the basic problem.

### DIFF
--- a/capstone/Dockerfile
+++ b/capstone/Dockerfile
@@ -27,7 +27,7 @@ RUN apt-get update \
     && apt-get install -y redis-server postgresql-client libtiff-tools pdftk \
     && apt-get install -y librocksdb5.17 librocksdb-dev libsnappy-dev zlib1g-dev libbz2-dev libgflags-dev liblz4-dev rocksdb-tools \
     && apt-get install -y libx11-xcb1 libxtst6 libgtk-3-0 libnss3 \
-    && apt-get -t buster-backports install -y libhyperscan-dev \
+    && PKG=$(( ${BUILDARCH} == 'arm64' ? 'libvectorscan-dev' : 'libhyperscan-dev' )) apt-get -t buster-backports install -y ${PKG} \
     && apt-get install -y tidy
 
 # pip
--- a/capstone/Dockerfile
+++ b/capstone/Dockerfile
@@ -1,6 +1,9 @@
 FROM python:3.7-buster
 ENV PYTHONUNBUFFERED 1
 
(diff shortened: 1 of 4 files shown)


bensteinberg:
This may be a dealbreaker for running natively on M1 machines: 

> Unable to locate package libhyperscan-dev

Debian's package page points out that 

>  libhyperscan only runs on processors with at least SSSE3 support. 

See the [Wikipedia article on SSSE3](https://en.wikipedia.org/wiki/SSSE3).

bensteinberg:
Blockers here include not just hyperscan, but pymupdf and pynacl as well. Further, trying to build the image on an M1 with `platform: linux/amd64` gets
```
=> => # Configuring libhyperscan5
 => => # -------------------------
 => => # This CPU lacks support for the Supplemental Streaming SIMD Extensions 3 (SSSE3)
 => => # instruction set that is required to execute programs linked against hyperscan.
 => => # Really install package? [yes/no]
 ```
 which suggests that dev on an M1 won't be possible without a larger effort.

jcushman:
pymupdf and pynacl both at least are build problems rather than unsupported-by-policy problems, which means they may work from package updates or somesuch.

As a general approach, is it viable to get the tests running, forcing unsupported dependencies to install, and see what % fail? I think a small % would actually require pymupdf or hyperscan, though it might require deferring package imports. If we can get most tests passing I might just mark them failing on ARM and wait for the world to catch up.

bensteinberg:
Thanks. I got `libhyperscan-dev` to install in 6541c50 and all tests passed locally.... Maybe we're not exercising the parts of hyperscan that use SSSE3?

(First `pytest` run locally took 10:20, second run took 6:45.)

bensteinberg:
By the way, the bump of two minor versions in the base Postgres image is because I upgraded prod one step without a matching PR. We have 11.13 here, 11.14 on prod, and I'm proposing 11.15 in this PR.

https://www.postgresql.org/docs/release/11.14/

https://www.postgresql.org/docs/release/11.15/

bensteinberg:
(Note that the original conception of this PR, allowing developers to run natively on M1 Macs, has changed to "allowing devs with M1 Macs to run at all" -- the original matching change in harvard-lil/docker-compose-update-action has also been reverted, so we're no longer attempting arm64 builds, but we can try again later.)

PR #2050, opened 2022-02-16
Package updates

bensteinberg:


bensteinberg:
1485541 is needed due to a change in tokenizing in nltk 3.6.6, I believe: https://github.com/nltk/nltk/blob/develop/ChangeLog

PR #2084, opened 2022-07-26
end-to-end containerized tests playwright

sabzo:
Please refer to #2085 on how this PR may fit.

### DIFF
--- a/capstone/config/settings/settings_base.py
+++ b/capstone/config/settings/settings_base.py
@@ -9,7 +9,7 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__fil
 SERVICES_DIR = os.path.join(os.path.dirname(BASE_DIR), 'services')
 
 
-ALLOWED_HOSTS = ['localhost', '127.0.0.1', '[::1]', '.test']
+ALLOWED_HOSTS = ['localhost', '127.0.0.1', '[::1]', '.test', '.web']
 
 ADMINS = [('Caselaw Access Project', 'info@case.law')]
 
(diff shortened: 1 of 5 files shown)


capstone/config/settings/settings_base.py
```diff
@@ -9,7 +9,7 @@
 SERVICES_DIR = os.path.join(os.path.dirname(BASE_DIR), 'services')
 
 
-ALLOWED_HOSTS = ['localhost', '127.0.0.1', '[::1]', '.test']
+ALLOWED_HOSTS = ['localhost', '127.0.0.1', '[::1]', '.test', '.web']
```
sabzo, line 12: Designed to allow http calls made in the format http://web:8000/ etc.,

rebeccacremona:
I'm taking the liberty of converting this to a draft; toggle back when/if you think it wants merging!

capstone/config/settings/settings_base.py
rebeccacremona: To work properly locally, CAP needs to be served from several [specific hosts](https://github.com/harvard-lil/capstone#user-content-hosts-setup). A person can tweak `docker-compose.yml` so that, container to container, you can use those same hosts, instead of the `web` service name. You have to set aliases `network` like [this](https://github.com/harvard-lil/perma/blob/develop/docker-compose.yml#L60-L65) and `extra_hosts` (the equivalent of /etc/hosts) like [this](https://github.com/harvard-lil/perma/blob/develop/docker-compose.yml#L52-L55).

Then, you can, for instance, `curl http://cap.test:8000` from your laptop AND from inside any of the containers, e.g. `docker compose exec db curl http://cap.test:8000` (presuming curl is installed in the db container lol)

PR #2090, opened 2022-08-09
Feature/end to end tests single container

sabzo:
WIP


capstone/playwright/test_homepage.py
```diff
@@ -0,0 +1,12 @@
+from playwright.sync_api import Page, expect
+
+url = "http://case.test:8000"
```
sabzo: Ideally I feel as if this should be defined as a constant somewhere in the codebase, but not sure where

capstone/test_data/test_fixtures/fixtures.py
```diff
@@ -392,3 +392,68 @@ def elasticsearch(request, settings):
...
+  'nj',
+  'nm',
+  'n-mar-i',
+  'ny',
+  'ohio',
+  'okla',
+  'or',
+  'pa',
+  'pr',
+  #'regional',
```
sabzo: TODO: Remove comments from list

capstone/playwright/test_homepage.py
rebeccacremona: Ah! Have you been running this against your local development server (spun up via `fab run` in another terminal tab), rather than against the special Django live test server that you can spin up from the test? E.g.

```
def test_my_thing(live_server, page):
    page.goto(f"{live_server.url}/route-in-question")
```

I think that's what you want here :-) Happy to chat about this offline if preferred!

In Perma, we ended up specifying that the test server should run on a particular port (`8000`), and set things up so that you could visit it with named hostnames instead of just localhost: though we can totally do something similar here, by default the live_server.url will look something like "http://localhost:xxxx", with the port randomly chosen.

rebeccacremona: That's why the [tests are failing](https://github.com/harvard-lil/capstone/runs/7754196798?check_suite_focus=true) in Github Actions: it doesn't have the development server running in the background :-)

A person can indeed run the development server in the background. For example, here's [an old version of Perma](https://github.com/harvard-lil/perma/tree/eff4c18a2e9d3c042667812f9f87bca810bf1238). With this version of Perma's test suite, we ran the end-to-end tests from their [own Docker container](https://github.com/harvard-lil/perma/tree/eff4c18a2e9d3c042667812f9f87bca810bf1238/playwright), instead of [from inside the regular test suite](https://github.com/harvard-lil/perma/tree/develop/perma_web/functional_tests) like we do now. In that case, the Django test server was not available, so we had to ask Github Actions to [spin up the development server using `fab run`](https://github.com/harvard-lil/perma/blob/eff4c18a2e9d3c042667812f9f87bca810bf1238/.github/workflows/tests.yml#L74) before the end-to-end/functional tests were run.

But, that has lots of disadvantages, as we've been chatting about. So, we moved the tests into the main test suite in the `web` container, like we've been chatting about.

Since this PR is following the second pattern, the one where the end-to-end tests are going to live inside the main test suite in the `web` container, though, we don't need to: we can just use `live_server`, and get all its advantages.

sabzo:
hold of peer review as still incomplete

capstone/playwright/test_homepage.py
rebeccacremona: I think maybe let's look at this again together in the python `debugger` on your computer some time, and we can mess around and look at the difference between the two setups. Lemme know if that's of interest!

sabzo:
Playwright isn't able to automatically click on <a> tags for states that is part of the svg states map. The svg intercepts "pointer" events when trying to click the link. This was needed to test if clicking a state graphic in the svg leads to the correct page.


E         <svg role="img" tabindex="" version="1.1" viewBox="0 …>…</svg> intercepts pointer events                                                                
E       retrying click action, attempt #60  

image

goal maybe to combine selectors to get more specific by adding the "polygon" tag.

lizadaly (approved):
I see that after the last couple commits the playwright tests still ran successfully, so I vote that this is ready to go!

PR #2065, opened 2022-05-03
Migrate from webpack to vite

jcushman:
Here's a first cut at migrating from webpack to vite. It does have the potential to break any given JS file, and I only got partway through checking all of the entrypoints, so this would need more testing to merge. Would also want to check how it ends up integrating with whitenoise.

PR #2095, opened 2022-09-14
Update two JS dependencies

bensteinberg:
https://github.com/vitejs/vite/blob/v2.9.15/packages/vite/CHANGELOG.md

https://github.com/moment/moment/blob/develop/CHANGELOG.md

### DIFF
--- a/capstone/package.json
+++ b/capstone/package.json
@@ -24,7 +24,7 @@
     "split-grid": "^1.0.9",
     "tone": "^14.8.39",
     "url-polyfill": "^1.1.3",
-    "vite": "^2.9.5",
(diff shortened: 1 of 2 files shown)


rebeccacremona (approved):
👍 Neat, I didn't know about "resolutions"

bensteinberg:
> 👍 Neat, I didn't know about "resolutions"

I think it would be possible not to include the resolution in the commit -- that is, you can add it, run `yarn`, then remove it once `yarn.lock` is in the right shape. By leaving it in, you end up over time with a pretty long list. In #2065 @jcushman removed such a list.

PR #2097, opened 2022-10-05
Point version update to Django

bensteinberg:
https://docs.djangoproject.com/en/dev/releases/3.2.16/


matteocargnelutti (approved):

bensteinberg:
Tests did not run because GH was having trouble yesterday, and they still haven't run; I'm going to merge this in the expectation that the tests will run and pass.

PR #2111, opened 2022-12-19
Handle the absence of a PDF

### LINKED ISSUE #2109: Return 404 for absent PDF, not 500
For cases that have no PDF file, like [this one](https://cite.case.law/s-ct/137/810/12609670/) from Fastcase, an attempt to load a PDF file produces an error:

> TypeError at /pdf/12609670/BNSF Ry. Co. v. Tyrrell.pdf
> unsupported operand type(s) for +: 'NoneType' and 'str'

It should probably produce a 404 instead.

bensteinberg:
Fixes #2109

### DIFF
--- a/capstone/cite/views.py
+++ b/capstone/cite/views.py
@@ -173,7 +173,7 @@ def volume(request, series_slug, volume_number_slug):
     })
 
 
-def case_pdf(request, case_id):
+def case_pdf(request, case_id, pdf_name):
     """
         Return the PDF for a case. This wraps citation() so that all rules about quotas and anonymous users can be
         applied before we return the case.
--- a/capstone/cite/views.py
+++ b/capstone/cite/views.py
@@ -173,7 +173,7 @@ def volume(request, series_slug, volume_number_slug):
     })
 
 
-def case_pdf(request, case_id, pdf_name):
+def case_pdf(request, case_id):
     """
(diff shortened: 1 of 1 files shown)


capstone/cite/views.py
```diff
@@ -179,12 +179,19 @@ def case_pdf(request, case_id, pdf_name):
...
+        pk=case_id
+    )
     pdf_url = case.get_pdf_url(with_host=False)
     if iri_to_uri(request.path) != pdf_url and not request.GET.get('redirect'):
-        return HttpResponseRedirect(pdf_url+"?redirect=1")
+        try:
+            return HttpResponseRedirect(pdf_url+"?redirect=1")
+        except TypeError:
+            # there is no pdf_url
+            raise Http404
```
lizadaly: Probably doesn't matter much here but I think you'll get more predictable behavior explicitly checking for whether this variable is `None` and raising earlier, rather than inferring the problem from `TypeError`.

lizadaly (approved):

capstone/cite/views.py
bensteinberg: Thanks. I wasn't sure what was most idiomatic/most directly applicable to the error in question. I think I'll check for `None`.

bensteinberg:
[8829546](https://github.com/harvard-lil/capstone/pull/2111/commits/8829546c157e0c5b50d946a8a3368ac878fa7272) would have been a nice moment for the walrus operator.

PR #2128, opened 2023-03-10
Fix PDF unredaction

jcushman:
Fix an issue where volumes being unredacted resulted in a new unredacted PDF appearing beside the original, with a modified filename like "/mnt/storage/prod/downloads/PDFs/open/Arkansas/Ark./165 Ark_WYVf0Q5.pdf", rather than replacing the redacted version. Existing instances of this problem will have to fixed with a separate script.

### DIFF
--- a/capstone/capdb/models.py
+++ b/capstone/capdb/models.py
@@ -4,6 +4,7 @@ import json
 import math
 import re
 import random
+import shutil
 from contextlib import contextmanager
 import struct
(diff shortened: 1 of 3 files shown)


lizadaly (approved):

lizadaly:
The failing test seems like a side effect of this change (which suggests maybe that test is too broadly scoped).

Issue #1974, opened 2021-08-11
Analysis explainer '?' may be redirecting incorrectly


bjohnson747:
when i click the Analysis '?' on the left side, below:
<img width="912" alt="Screen Shot 2021-08-04 at 8 55 00 PM" src="https://user-images.githubusercontent.com/12850389/129093179-ea473842-08f9-456f-9c98-8f09ca129258.png">

 it doesn't jump me down to any specific section in the API doc:
<img width="858" alt="Screen Shot 2021-08-04 at 8 55 17 PM" src="https://user-images.githubusercontent.com/12850389/129093187-a69bf259-55e9-482b-b58b-39330736b6e2.png">

ChefAndy:
This could point to `{% docs_url 'data_formats' %}#analysis-fields` but that explanation probably isn't great for folks surfing onto a case from google. We should have a friendlier little blurb.

kilbergr:
https://github.com/harvard-lil/capstone/pull/2129

PR #2129, opened 2023-03-22
Updates some broken urls and pulls relevant explanation higher

kilbergr:
This PR fixes a [broken link on the analysis explainer](https://github.com/harvard-lil/capstone/issues/1974) and a couple of instances of a broken link for case filtering. I pulled the info explaining what the analysis stats are up as well so that unfamiliar users would read that before the stats break down.


matteocargnelutti (approved):
🚀

Issue #1920, opened 2021-06-25
check if case ID volumes are out of scope in CAP


4ndygu:
https://github.com/harvard-lil/capstone/blob/61971f22cf2d35ee4ae38e21a9f67c3f1a45756c/capstone/fabfile.py#L1302 checks whether extracted citations include a target_case is in_scope, but not necessarily whether `case.volume.out_of_scope` is false. As a result, certain cases inaccessible from the API are persisted in the citation graph. 

![image](https://storage.googleapis.com/andyresume23487623846283764/test.png)

Thanks for pointing out the root cause @jcushman ! can also test myself after hours.

jcushman:
I think what we want to do here is add a denormalized `volume_out_of_scope` field to `CaseMetadata` that contains a copy of `case.volume.out_of_scope`, so we can efficiently update `case.in_scope` without loading the `case.volume` foreign key. Once that's done we can just check `case.in_scope` and not have to worry about the volume.

kilbergr:
@jcushman has this db been normalized?

kilbergr:
@rebeccacremona kindly pointed me to this: https://github.com/harvard-lil/capstone/pull/268 which makes it seem like the db was normalized and uses denormalized fields in this specific way.

kilbergr:
I have been doing some digging and it looks like with the introduction of ES, the denormalization functionality was removed because it was no longer necessary: https://github.com/harvard-lil/capstone/commit/27b4f10fc34bc0c2313dc1116c1ac40832e46331
I'm digging to figure out why ES functionality had that effect, but if anyone wants to point me somewhere feel free.

kilbergr:
Update:
Adding [this level](https://github.com/harvard-lil/capstone/pull/268) of scaffolding for a single denormalized field seemed like a lot of work. I ran an experiment where I added an inner join to the query in question to make sure cases from volumes where `out_of_scope` was false.

```
select
            cited_by.id
        from
            capdb_extractedcitation ec
            inner join
                capdb_casemetadata target_case on ec.target_case_id = target_case.id
            inner join
                capdb_casemetadata cited_by on ec.cited_by_id = cited_by.id
            inner join
                capdb_volumemetadata target_volume on target_case.volume_id = target_volume.barcode
        where 
            cited_by.in_scope is true
            and target_case.in_scope is true
            and target_volume.out_of_scope is false
            and cited_by.decision_date_original >= target_case.decision_date_original
            and cited_by.id != target_case.id
[...]

PR #2130, opened 2023-03-27
Adds a check for out_of_scope volumes to citation extraction

kilbergr:
This PR addresses the [check if volumes are out of scope for cases when exporting citation graph](https://github.com/harvard-lil/capstone/issues/1920) issue.

It adds another inner join to check the volume `out_of_scope` field. We've tested how much this slows the query added here [in that issue](https://github.com/harvard-lil/capstone/issues/1920#issuecomment-1485687774) and this seems like a reasonable fix.

The test adds a case in an out of scope volume with a citation. Like the `future_cite` or `cite_not_in_cap`, it is not included in the exported citation graph.

Comments welcome

### DIFF
--- a/capstone/capdb/tests/test_tasks.py
+++ b/capstone/capdb/tests/test_tasks.py
@@ -414,15 +414,16 @@ def test_run_text_analysis(reset_sequences, case):
 
 
 @pytest.mark.django_db(databases=['capdb'])
-def test_export_citation_graph(case_factory, tmpdir, elasticsearch, citation_factory, jurisdiction_factory):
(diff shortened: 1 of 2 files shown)


kilbergr:
@jcushman you may be interested in this. See conversation in issue: https://github.com/harvard-lil/capstone/issues/1920

jcushman:
Nice! This looks like a good fix to me.

rebeccacremona (approved):

kilbergr:
@jcushman -- we need to run the command `fab export_citation_graph` independently for this change to take effect. Do I need to know anything I might not know before merging and asking Ben to push and run on the prod server?

Issue #1981, opened 2021-08-23
Fix boolean logic across opinions


jcushman:
Switching the cases index to nested elasticsearch documents per opinion had the effect of changing the boolean logic for search queries.

If you do `?search=foo bar`, that should mean "find cases where one of the metadata fields or one of opinions contains 'foo', and one contains 'bar'", but it now means "find cases where one of the metadata fields or one of opinions contains both 'foo' and 'bar'."

The fix is probably to use the `copy_to` attribute on the fields we want to search and copy them to a single elasticsearch column, which would also make the search a little faster. But this needs research because it would also increase the index size, potentially too much for the server.

kilbergr:
Comprehension check--
You want the `?search=foo bar` query to return cases where either:
1. one of the metadata fields contains 'foo' and one of opinions contains 'bar'
2. one of the opinions contains 'foo' and one of metadata fields contains 'bar'

Is that right @jcushman ?

jcushman:
Good question! Starting from a user experience perspective, the goal would be that if I do an "and" search for foo and bar, such as `?search=foo bar`, and a given case page like cite.case.law/mass/1/1/ shows both the words foo and bar, then the case should match.

Right now elasticsearch is requiring that ANDed terms all appear in the same section of the page, so it'll wrongly fail to return /mass/1/1/ if foo happens to be in one part of the page while bar is in another.

kilbergr:
Further clarification from Slack https://hlslil.slack.com/archives/C0JS2FR3J/p1679583130799379
This was not referencing the `CaseMetadata` model, but instead metadata (nested data) in the [CaseDocument model](https://github.com/harvard-lil/capstone/blob/64c3dc16bb6083398a2a32938cf179c27d942657/capstone/capapi/filters.py#L501).
So the issue is: if `foo` shows up in one of those fields and `bar` shows up in an opinion for a case, will that case be returned for the `?search=foo bar` query.

kilbergr:
Process for now: close this ticket because we have [a fix](https://github.com/harvard-lil/capstone/pull/2132). Leave the fix open until [the deployment](https://github.com/harvard-lil/capstone/issues/2133) is complete. Have we created a recursion issue creation loop? Only time will tellllllllllll.

PR #2132, opened 2023-04-06
Fix boolean search

kilbergr:
This PR addresses the [issue around boolean search failure](https://github.com/harvard-lil/capstone/issues/1981) due to ES not searching across multiple fields when using the AND keyword.
To fix, I implemented `copy_to` to grab the content of all the fields we wanted to search into a single field that could be searched.
This is what happens when you search `"Argued before BELL" AND "rationale of Gardner"` on the live site:
Screenshot 2023-04-06 at 3 47 47 PM
That's because the first phrase is in the `head_matter` field and the second is in `casebody_data.opinions.text`. 

This is what happens when you run the same search with this change:
Screenshot 2023-04-06 at 3 48 39 PM

Deploying this will require running `fab rebuild_search_index` on prod, so we will have to plan when that is appropriate.

### DIFF
--- a/capstone/capapi/documents.py
+++ b/capstone/capapi/documents.py
@@ -76,16 +76,17 @@ class RawSearch(Search):
 class CaseDocument(Document):
     # IMPORTANT: If you change what values are indexed here, also change the "CaseLastUpdate triggers"
     # section in set_up_postgres.py to keep Elasticsearch updated.
-    name_abbreviation = SuggestField(analyzer='english')
-    name = fields.TextField(index_phrases=True, analyzer='english')
+    name_abbreviation = SuggestField(analyzer='english', copy_to="multi_field_content")
+    name = fields.TextField(index_phrases=True, analyzer='english', copy_to="multi_field_content")
     frontend_url = fields.KeywordField()
     frontend_pdf_url = fields.KeywordField()
     last_page = fields.KeywordField()
     first_page = fields.KeywordField()
     decision_date_original = fields.KeywordField()
     docket_numbers = fields.TextField(multi=True)
-    docket_number = fields.TextField()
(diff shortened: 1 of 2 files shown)


bensteinberg (approved):
LGTM! But I will add @jcushman as a reviewer as well.

jcushman:
LGTM2! Note that this will increase the index size, so disk space on prod could be a question; not sure if there's a way to estimate.

kilbergr:
That is a good question. Let me try to find out.
Also @bensteinberg I wondered whether we'd want to push the piece that is necessary for the reindex, run the reindex, then push the change to search. Idk if that's overkill...

kilbergr:
@jcushman can you clarify for me: what is our plan should the index size outgrow disk space? Will we increase the disk space?

mdellabitta:
> @jcushman can you clarify for me: what is our plan should the index size outgrow disk space? Will we increase the disk space?

That's my expectation. But Ben would have to be in charge of how that happens, and LOE.

jcushman (approved):

capstone/capapi/documents.py
```diff
@@ -76,16 +76,17 @@ def execute(self, ignore_cache=False):
 class CaseDocument(Document):
     # IMPORTANT: If you change what values are indexed here, also change the "CaseLastUpdate triggers"
     # section in set_up_postgres.py to keep Elasticsearch updated.
-    name_abbreviation = SuggestField(analyzer='english')
-    name = fields.TextField(index_phrases=True, analyzer='english')
+    name_abbreviation = SuggestField(analyzer='english', copy_to="multi_field_content")
```
SFQUEEN94, line 79: Analysis

SFQUEEN94 (approved):
Approved

PR #2149, opened 2023-06-09
Adds an index block to pages with meta tags

kilbergr:
This work corresponds to issue 94 [on Linear](https://linear.app/harvard-lil/issue/ENG-94/cap-remove-cases-from-google-indexsearch-platforms)

It is part of the effort to make deindexing permanent for CAP pages. This is what will stop our site from getting re-indexed. The process is [described here](https://developers.google.com/search/docs/crawling-indexing/remove-information).

main_base.html is the only page that is used throughout the site as the template all other templates build from, so I think that's actually the only one really necessary here.


mdellabitta (approved):
🚀

PR #2151, opened 2023-06-27
Add bounding rectangles from PDF to each paragraph in exported HTML

jcushman:
Add the data-blocks attribute to exported HTML, listing the (arbitrary assigned by vendor) block ID, PDF page order, and bounding box for each rectangle in the PDF that contributed to the paragraph. Example output:

```
<p id="b415-10" data-blocks='[["BL_415.10",415,[249,2172,1145,146]],["BL_416.1",416,[314,411,1150,756]]]'>Chief Justice Fuller in <em>United States v. Lacher ...
```

Bex will pick up stewarding this PR from here; once we're happy with it, we'd like to run `fab refresh_case_body_cache` on prod and see if all the volumes work.


kilbergr (approved):
🥇

PR #2158, opened 2023-07-26
Various dependency updates

bensteinberg:
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/release-notes-7.17.12.html

(@kilbergr note the boto3 update)


matteocargnelutti (approved):
Ah yes yes I have reviewed the hashes and they look great hm 🧐

bensteinberg:
Initial build was failing on rocksdb compilation -- I've pushed a Cython change that may help, as well as a Playwright update that I forgot to commit before 🙄

bensteinberg:
I see paths including Cython 3 in the failing build, so I'm trying to unpin Cython now.

bensteinberg:
I believe what's happening here is that Cython 3 was released a couple of weeks ago, and the python-rocksdb setup.py, calling for `'Cython>=0.20'`, is now getting the new version, not what we previously had pinned. I'm going to try to force the build to use an older Cython, or conceivably fork python-rocksdb. Stay tuned. I think this is a blocker for any dependency updates in requirements.txt.

kilbergr (approved):
Thank you for the updates!

PR #2161, opened 2023-08-01
Rek refactor s3 convert

kilbergr:
This is the PR for refactoring changes @jcushman requested, so I'm going to tag him as a reviewer. 
First, reduce the number of db queries.
Second, change the way the cases metadata is being generated by having it generated based on what is first generated for the cases themselves.
Third, change the way file names are created. Instead, order the cases by id and use a comparison of the previous case's first page to determine whether a file should be later in a series.
Fourth, change the way we check whether files exist. Compare hashes and names of files gleaned from S3 to the current hashes and file names.

The only commit I've made: https://github.com/harvard-lil/capstone/pull/2161/commits/4ca3d1672f809096d051096a0d86bd9a9d8791e4
The only file I've changed is https://github.com/harvard-lil/capstone/pull/2161/files#diff-b957af52fc247bd70b117433d1fb7fbdf64bdbff62ce5259934ffa1323f6c4d2
I'm getting some side effects from using the experimental rather than the develop branch--I merged the current lil develop into my forked experimental before making this PR and am now trying to merge those into the experimental with my actual change.


capstone/scripts/convert_s3.py
```diff
@@ -76,9 +75,10 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
         return
 
     # determine prefix based on a case's frontend_url
-    sample_case_obj = CaseMetadata.objects.select_related().filter(
-        reporter=reporter_id
+    sample_case_obj = list(
+        reporter.case_metadatas.select_related().order_by("case_id")
     )[0]
+
```
jcushman: Can you just use `reporter.short_name_slug` rather than look at cases? I'm not sure how these two techniques compare. You should check in either case if these are unique -- do more than one reporter have the same one? 

If you keep this technique, `list(query)[0]` is less efficient and clear than `query.first()`, which only fetches one row rather than fetching thousands and keeping the first. You also don't want `select_related()` here since you aren't using any related objects on the case.

capstone/scripts/convert_s3.py
```diff
@@ -89,123 +89,135 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
     put_reporter_metadata(bucket, reporter, reporter_prefix)
 
     # get volumes in reporter
-    volumes = VolumeMetadata.objects.select_related().filter(reporter=reporter_id)
-    # export volume metadata/cases
-    export_cases_by_volume(volumes, bucket, redacted)
+    volumes = list(reporter.volumes.select_related())
```
jcushman: I think this can just be `for volume in reporter.volumes` -- you're not touching each volume more than once, so `list` isn't necessary, and I don't see where `select_related` is being used.

jcushman: Oh but we also need some filtering on which volumes and reporters are in scope, right?

capstone/scripts/convert_s3.py
```diff
@@ -89,123 +89,135 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
...
-                    file.write(json.dumps(serializer.data).encode("utf-8") + b"\n")
-                file.flush()
-                # not closing with loop so I can continue using file for upload
-                hash_and_upload(file, dest_bucket, key, "application/jsonl")
+    cases_by_id = {case.pk: case for case in cases}
+    frontend_url = cases[0].frontend_url
 
-        # copies each volume PDF to new location if it doesn't already exist
-        copy_volume_pdf(volume, volume_prefix, dest_bucket, redacted)
+    if len(cases) == 0:
```
jcushman, line 120: The previous line assumes len(cases) > 0.

capstone/scripts/convert_s3.py
```diff
@@ -89,123 +89,135 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
...
+            else:
+                file.write(case_contents)
+                file.flush()
+                hash_and_upload(
+                    file,
+                    dest_bucket,
+                    f"{cases_key}{case_file_name}",
+                    "application/jsonl",
+                )
+
```
jcushman, line 171: Here we have to delete anything left in `s3_contents_hash` from s3; otherwise making a change to the export format and rerunning will result in double files.

capstone/scripts/convert_s3.py
```diff
@@ -89,123 +89,135 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
...
+            file,
+            dest_bucket,
+            f"{volume_prefix}/CasesMetadata.jsonl",
+            "application/jsonl",
+        )
+
+    # copies each volume PDF to new location if it doesn't already exist
+    copy_volume_pdf(volume, volume_prefix, dest_bucket, redacted)
+
+    # export each case in the volume
```
jcushman: Remove comment?

capstone/scripts/convert_s3.py
```diff
@@ -89,123 +89,135 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
...
         accepted_renderer=None,
     )
+    # fake Request object used for serializing cases with DRF's serializer
+    vars["fake_request"] = namedtuple("Request", ["query_params", "accepted_renderer"])(
+        query_params={"body_format": "text"},
+        accepted_renderer=None,
+    )
+
+    # create a metadata contents string to append case metadata content
+    metadata_contents = b""
```
jcushman: As a tiny nitpick, I think python is smoother if you keep everything internally in strings, and just do `.encode("utf-8")` right at the end when you call `hashlib.sha256` or `file.write`. In this case you can also do `tempfile.NamedTemporaryFile(mode='wt')` to specify opening the file in text mode, and then I don't think you'll need `.encode` at all.

capstone/scripts/convert_s3.py
```diff
@@ -89,123 +89,135 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
...
+                file.write(case_contents)
+                file.flush()
+                hash_and_upload(
+                    file,
+                    dest_bucket,
+                    f"{cases_key}{case_file_name}",
+                    "application/jsonl",
+                )
+
+    with tempfile.NamedTemporaryFile() as file:
```
jcushman: None of these tempfiles are helping you, since `hash_and_upload` just reads the file back into a string. I would update `hash_and_upload` to take a `str` and remove these.

capstone/scripts/convert_s3.py
```diff
@@ -356,66 +354,29 @@ def copy_volume_pdf(
...
-    return case_name
+        raise Exception(f"Cannot list objects {bucket}/{key}: %s" % err)
+    if "Contents" not in response:
+        return s3_contents_hash
+    else:
+        for case in response["Contents"]:
+            case_file_name = case["Key"].split("/")[-1]
+            # Get the object's metadata
+            try:
+                response = s3_client.get_object_attributes(
```
jcushman, line 360: The checksum should be in the original list response https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_objects_v2.html , so we shouldn't need an additional API call per file.

jcushman (approved):
Great progress! I added some requested changes. I'd suggest pulling as much as possible into develop, assuming none of it interferes with prod since it's mostly in convert_s3.py, and iterating from there.

capstone/scripts/convert_s3.py
kilbergr: There aren't out of scope reporters, afaict. There are those with no volume_count, but then we'd still want the reporter metadata file I believe.

capstone/scripts/convert_s3.py
kilbergr: Oof good catch--they're not unique.
These all have 2 reporters with the same slug, which we're using to delineate which reporter is which in our S3 buckets. We could do something similar as we do with the cases and make a 1 and 2 version, but that approach makes less sense to me in this case because it's not that they're actually ordered that way. There might be an argument for using the `analyst_start_year` column to order them, but there are a couple of drawbacks: that column can be null or blank, and some reporters seem to cover the same time period (i.e. one begins in 1941 and goes to 1942, another with the same slug goes from 1941 to 1943). 
Is there a different approach that makes more sense?

The repeats:

```
 short_name_slug | count
-----------------+-------
 bur             |     2
 cal             |     2
 conn-super-ct   |     2
 ct-cl           |     2
 dallam          |     2
 davis-l-ct-cas  |     2
 haw             |     2
 ind-app         |     2
 la              |     2
 lack-bar-r      |     2
 leg-gaz         |     2
 luz-leg-reg     |     2
 mass-app-div    |     2
 nys-2d          |     2
 va              |     2
```

capstone/scripts/convert_s3.py
jcushman: My guess is a reporter without volumes doesn't have any useful metadata either and we just exclude it.

capstone/scripts/convert_s3.py
jcushman: Can you take a look at each, like https://cite.case.law/bur/? My guess is in most cases one or both of these doesn't have volumes, and if we exclude those this goes away. The exception I saw in the first few was Ct. Cl., where one is West Virginia Court of Claims and one is US Court of Claims but both reporters properly have their cases cited as "1 Ct. Cl. 1" -- https://cite.case.law/ct-cl/1/. SO ... us-ct-cl and wv-ct-cl would make sense for the folder names, but I'm not sure if it would break anything to set reporter short name that way in the database or if we need a special case in the export script.

kilbergr: Proposal: new naming convention as slug-reporter_id (i.e. `ind-app-436` and `ind-app-698`)

capstone/scripts/convert_s3.py
kilbergr: I believe that the checksum algorithm is in the response, but the checksum is not. There is some chatter about the ETag being used for this, but according to the documentation:
“The ETag may or may not be an MD5 digest of the object data. Whether or not it is depends on how the object was created and how it is encrypted as described below”
Part of that is determined by size of the object, among other criteria (although I think it’s extremely unlikely we’ll hit the size limitation), so seems risky to use. That’s why I chose to do a second query. Does that seem misguided?
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_objects_v2.html

jcushman: Ah got it, that's annoying -- thanks for checking. I think if speed of recreating a volume became an issue, switching to ETags might be a good idea. But you've also convinced me I'm overoptimizing, so what you have makes sense to me.

capstone/scripts/convert_s3.py
kilbergr: This is a really good point and obviates the need to do anything else for this:
https://github.com/harvard-lil/capstone/pull/2161#discussion_r1286194995

capstone/scripts/convert_s3.py
kilbergr: See below: taken care of by using strings rather than tempfiles.

capstone/scripts/convert_s3.py
kilbergr: I'm confused here. Here's the situation: 
Right now we query S3 for files and hashes and hold in a dict.
If a case's hash is the same as an S3 file hash and the names are the same, pop it from the dict and don't upload.
Otherwise, write a new S3 file from the case contents.

If I'm reading you correctly, what I think you're pointing out is:
When there's already a case of that name on S3, but the file hashes are different, we should delete that file from S3 and write a new one.

So that would look to me like:
if hashes are the same and file names are the same, pop from dict and don't upload (continue)
else if file names are the same but hashes are not the same, delete from S3, and upload new file
else upload new file.

Is that the correct interpretation?

kilbergr: Michael helped get to a different understanding. We now think you're saying: pop if the name and file contents are the same, if they're not upload and then pop, and at the end if you have anything else left, delete those.

jcushman: How about something like this?

```
s3_existing_object_hashes = {'key': 'hash', ...}  # map keys to hashes

for item in cases_search.scan():
    # do upload if file changed
    dest_key = f"{cases_key}{case_file_name}"
    s3_contents_hash = s3_existing_object_hashes.pop(dest_key)
    if s3_contents_hash is None or s3_contents_hash != case_contents_hash:
        ... # do upload

for extra_s3_object in s3_existing_object_hashes:
    ... # delete extra_s3_object
```

We push the contents for any file that doesn't have the same contents already existing at that key. And then if any keys aren't encountered at all, we know they're an extra file and need to be deleted.

kilbergr:
Reached out to @jcushman to ask for one last look.

capstone/scripts/convert_s3.py
kilbergr: Ticket created here: https://linear.app/harvard-lil/issue/ENG-283/determine-how-to-differentiate-among-reporters-with-same-slug

kilbergr:
@jcushman gives approval https://hlslil.slack.com/archives/C0JS2FR3J/p1692022594401659

PR #2163, opened 2023-08-14
Rek reporter slug

kilbergr:
This PR handles an issue Jack pointed out--that the reporters' slugs are not necessarily unique. After some digging, we identified that only two reporters that shared slugs actually have volumes/cases within them, so this code differentiates the identical slugs from each other. It also changes the way we check for reporters without volumes--we realized that volume_count did not reflect actual volume count, so now we're using `out_of_scope` instead, which seems like a heartier measure.

### DIFF
--- a/capstone/scripts/convert_s3.py
+++ b/capstone/scripts/convert_s3.py
@@ -77,7 +77,7 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
 
     # TODO: address reporters that share slug
     if reporter_id in reporter_slug_dict:
-        reporter_prefix = f"{reporter_slug_dict[reporter_id]}"
+        reporter_prefix = reporter_slug_dict[reporter_id]
     else:
         reporter_prefix = f"{reporter.short_name_slug}"
 
--- a/capstone/scripts/convert_s3.py
+++ b/capstone/scripts/convert_s3.py
@@ -65,7 +65,7 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
     reporter = Reporter.objects.get(pk=reporter_id)
 
     # Make sure there are volumes in the reporter
-    if reporter.volume_count is None or reporter.volume_count == "":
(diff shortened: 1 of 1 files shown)


capstone/scripts/convert_s3.py
```diff
@@ -213,6 +216,10 @@ def export_cases_by_volume(
 
 # Reporter-specific helper functions
 
+# Some reporters share a slug, so we have to differentiate with ids
+reporter_slug_dict = {"415":"us-ct-cl", "657":"wv-ct-cl",
```
jcushman, line 220: I think this won't work with string keys?

capstone/scripts/convert_s3.py
```diff
@@ -76,7 +76,10 @@ def export_cases_to_s3(redacted: bool, reporter_id: str) -> None:
         return
 
     # TODO: address reporters that share slug
-    reporter_prefix = f"{reporter.short_name_slug}"
+    if reporter_id in reporter_slug_dict:
+        reporter_prefix = f"{reporter_slug_dict[reporter_id]}"
+    else:
+        reporter_prefix = f"{reporter.short_name_slug}"
```
jcushman: (Since I had a substantive comment, I'll throw in that the `f""` here is redundant.)

jcushman (changes requested):

capstone/scripts/convert_s3.py
kilbergr: I tested it and it does work with string keys, also backed up by how on line 79, `type(reporter_id) <class 'str'>` so it doesn't work with the key as an int:

 https://github.com/harvard-lil/capstone/pull/2163/files#diff-b957af52fc247bd70b117433d1fb7fbdf64bdbff62ce5259934ffa1323f6c4d2R68

jcushman: oh I see, those IDs are from the API rather than database models -- carry on!

jcushman (approved):
LGTM!

PR #2165, opened 2023-08-17
updates all reporters command and creates a trial command

kilbergr:
This PR updates the all reporters command and creates a trial command so we can run this on prod with just a subset of reporters.
The reporters command had not been updated since some changes had been made to the bulk of the script that changed the way metadata for reporters and volumes were being saved. That was an oversight!
It's to address this issue: https://linear.app/harvard-lil/issue/ENG-12/test-s3-conversion-script-on-subset-of-prod-db

### DIFF
--- a/capstone/fabfile.py
+++ b/capstone/fabfile.py
@@ -458,15 +458,23 @@ def export_cases_to_s3():
     by reporter and volume.
     """
     redacted = True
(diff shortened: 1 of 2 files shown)


matteocargnelutti (approved):
🚀

mdellabitta (approved):

PR #2167, opened 2023-08-18
Rek add top level

kilbergr:
This PR addresses an issue that popped up when I realized I'd taken out the ReportersMetadata.jsonl and VolumesMetadata.jsonl that were supposed to live at the top level of the file structure. These contain metadata for all the reporters and all the volumes, respectively.

I ended up threading a tuple up with string containing the contents of each of these, but I'm not sure that is the best solution. I'm especially unsure about returning ("", "") for cases where the reporter has no volumes or cases. I just don't know whether that's the correct thing to do, so happy for advice on that.


kilbergr:
Adding Matteo because Ben, Becky, and Michael are all out next week.

matteocargnelutti (approved):
LGTM

PR #2169, opened 2023-09-06
Make export_cases_by_volume a shared task

bensteinberg:
I've tested this code with a local S3-alike minio container and `CELERY_TASK_ALWAYS_EAGER = False`. I have not included the minio setup here, but I can add it if that seems desirable.

Note that I had to change the volume argument to `export_cases_by_volume` from an object to a string, since the object is not JSON serializable. I suppose the excess queries could slow this down, but I don't think it will be a problem.

### DIFF
--- a/capstone/scripts/convert_s3.py
+++ b/capstone/scripts/convert_s3.py
@@ -5,12 +5,13 @@ import json
 import requests
 from botocore.exceptions import ClientError
 from collections import namedtuple
+from celery import group, shared_task
 
 from capapi.documents import CaseDocument
 from capapi.serializers import (
     ConvertNoLoginCaseDocumentSerializer,
 )
-from capdb.models import Reporter
+from capdb.models import Reporter, VolumeMetadata
 
 s3_client = boto3.client("s3")
 api_endpoint = "https://api.case.law/v1/"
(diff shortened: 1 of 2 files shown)


matteocargnelutti (approved):
LGTM

capstone/scripts/convert_s3.py
```diff
@@ -139,29 +140,41 @@ def export_cases_to_s3(bucket: str, redacted: bool, reporter_id: str) -> tuple:
...
-        reporter.volumes.exclude(volume_number=None)
-        .exclude(volume_number="")
-        .exclude(out_of_scope=True)
-    ):
-        # export volume metadata/cases
-        volume_metadata = export_cases_by_volume(
-            volume, reporter_prefix, bucket, redacted
+
+    job = group(
+        export_cases_by_volume.s(
```
rebeccacremona, line 145: TIL [`s`](https://docs.celeryq.dev/en/stable/reference/celery.app.task.html#celery.app.task.Task.s) and [signatures](https://docs.celeryq.dev/en/stable/userguide/canvas.html#signatures).

Pretty nice!

rebeccacremona (approved):

PR #2173, opened 2023-09-19
Correct fab task

bensteinberg:
Whoops; this fab task must not have caught up with scripts/convert_s3.py.


matteocargnelutti (approved):
Tests pending

PR #2177, opened 2023-10-12
Skip missing case IDs

bensteinberg:
This should resolve a KeyError that appears when the over-stuffed `cases` is compared with the more limited `cases_search_by_id`.


kilbergr (approved):
"overstuffed"
LGTM. Thank you!

PR #2179, opened 2023-10-23
Dependency updates

bensteinberg:
https://urllib3.readthedocs.io/en/latest/changelog.html#id10

https://docs.djangoproject.com/en/dev/releases/3.2.22/


tinykite (approved):
Looks reasonable to me!

PR #2184, opened 2023-12-04
CAP static export script updates

jcushman:
This is my first pass at updating the convert_s3.py script for exporting CAP data to cap static format:

- refactor the script to export to local disk instead of s3
- rename to export_cap_static.py instead of convert_s3.py
- update to use this workflow:
    - `fab export_cap_static_volumes` runs a celery task for each volume (or target volume) to export them to local disk. This step should take about 200 worker-hours, so less than a day given 16 workers.
    - `fab summarize_cap_static` gets run manually after all volumes are done, and writes reporter metadata and summary files
    - manually upload using something like rclone
    - a third command will be needed to move PDFs and CAPTAR files on S3 once the local files are uploaded -- Bex has `copy_volume_pdf` in the codebase that is mostly there
- format changes:
    - per slack discussion, just use .json rather than .jsonlines since we're not exporting any very large files
    - include html case files
    - add jurisdiction metadata files and reporter-level volume files
- Include unredacted volumes. We do this by temporarily redacting and then rolling back, which means the export task has to rely just on the database and not on the elasticsearch index.
    - Unredacted volumes are only exported if `settings.REDACTION_KEY` is set. My thought here is to use `CELERY_TASK_ROUTES` to route volume export tasks to a separate temporary celery worker that has that key set, but that might be over-complicated -- tbd with Ben.
- Add a test for the export task that stores expected output in test_data/cap_static, so it's easy to see what we expect the task to do and check for regressions.

Next stuff:

- This should be fine to pull now, if we're happy with the tests etc.
- I'm interested to hear what Dakota encounters in using these files to create the CAP static website, since that's a good test of what's in there.
- Let's try dumping some open jurisdictions and shipping to S3 to see how it goes.
- And then we can make the script to add the PDFs and CAPTARs.
- While I'm in here I'd like to throw in an export of the page structure for each volume, which is a small stretch goal but retains a bit of data we're otherwise not expressing anywhere.

Deployment notes:

- See the thing on a separate celery queue.


bensteinberg (approved):

bensteinberg (approved):
(Approved, that is, except I see tests are failing.)

PR #2185, opened 2023-12-20
Add unique case paths to static export

jcushman:
This PR tries to handle the tricky issue of how to find one case in the static file export from another.

**Problem:**
If you have a case like /us/123/cases/0012-01.json that cites to case id 4567, or a list of case IDs in /us/123/CaseMetadatas.json, it's hard to find the files that contain the data for those cases.

**Solution:**
Everywhere that we have cases referred to by ID, also refer to them by path. A path for a case is like "/us/123/0012-01", or just "0012-01" if the reporter and volume are already implicit.

So:
* In CaseMetadatas.json and single case json files, add a `"file_name": "0012-01"` field. This makes it possible to jump from CaseMetadatas.json to the associated json or html file, allowing the frontend to render the list of cases in a volume for example.
* In case json, when we link to a case this case cites with `"cites_to": [{"case_ids": [1,2,3...]}]`, add a field for each citation like `"case_paths":["/us/1/0001-01", "/us/2/0002-01", ...]` that parallels `"case_ids"`.
* In case html, when we link citations to their targets, edit them to point to case paths instead. This has a few options:
  * We know of no target case for the citation: link to `href="/citations/?q=123 Wtf. 456"`. (This is common for statutes and regulations).
  * We know of one or more target case for the citation: link to `href="/us/1/0012-01" data-case-paths="/us/1/0012-01,/us/1/0012-02"`.

**Deployment notes:**

Case paths are hard to calculate, because you have to add an increment for two cases starting on the same page of the same book, like `0123-01` and `0123-02`. So this PR adds a column in a migration to store that calculated string on CaseMetadata, and after deployment we should run `fab set_case_static_file_names` (which runs in a tqdm loop rather than in celery) before redoing `fab export_cap_static_volumes` for the non-redacted jurisdictions.


mdellabitta (approved):
Looks like a reasonable and minimal solution! 👍🏻

PR #2186, opened 2024-01-08
Update Elasticsearch

bensteinberg:
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/release-notes-7.17.16.html

### DIFF
--- a/capstone/docker-compose.yml
+++ b/capstone/docker-compose.yml
@@ -10,7 +10,7 @@ services:
     redis:
         image: redis:6.0.15
     elasticsearch:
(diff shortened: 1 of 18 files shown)


mdellabitta:
Looks like changes made to make CAP export happy blew this up, but I don't think it matters?

bensteinberg:
I'm looking into why the last updated rewrite isn't working.

bensteinberg:
last_updated is fine, the problem is with date_added

rebeccacremona:
Ah, nice catch! In case others in the future are looking: that 23-12-05 date is simply the last time these files were generated
image

rebeccacremona (approved):
LGTM!

Issue #2194, opened 2024-02-05
Errors in cap-static export


bensteinberg:
I see a few different errors in the export logs today; I think they were not introduced in #2192, the proximate reason for re-running the export, but I no longer have the logs from the previous export. FWIW these errors appear to have killed the export process early, which I think did not happen before; the redacted export from this run only got to 6.4G, which is about half what it should be, I believe.

The errors I see look like

```
[2024-02-05 15:19:00,087: ERROR/ForkPoolWorker-3] Task scripts.export_cap_static.export_cases_by_volume[69011773-07dd-4cdf-8918-efbf0babfb36] raised unexpected: TypeError('sequence item 1: expected str instance, NoneType found')
Traceback (most recent call last):
  File "/home/capstone/capstone-prod/lib/python3.7/site-packages/celery/app/trace.py", line 385, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/capstone/capstone-prod/lib/python3.7/site-packages/celery/app/trace.py", line 648, in __protected_call__
    return self.run(*args, **kwargs)
  File "/usr/local/share/capstone-prod/capstone/scripts/export_cap_static.py", line 109, in export_cases_by_volume
    export_volume(volume, dest_dir / "redacted")
  File "/usr/local/share/capstone-prod/capstone/scripts/export_cap_static.py", line 241, in export_volume
    el.attrib["data-case-paths"] = ",".join(el_case_paths)
TypeError: sequence item 1: expected str instance, NoneType found
```
and
```
[2024-02-05 15:19:30,483: ERROR/ForkPoolWorker-18] Task scripts.export_cap_static.export_cases_by_volume[60ad3c4e-0104-4dec-b5bb-4edb5862a1f2] raised unexpected: TypeError("Argument must be bytes or unicode, got 'NoneType'")
Traceback (most recent call last):
  File "/home/capstone/capstone-prod/lib/python3.7/site-packages/celery/app/trace.py", line 385, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/capstone/capstone-prod/lib/python3.7/site-packages/celery/app/trace.py", line 648, in __protected_call__
    return self.run(*args, **kwargs)
  File "/usr/local/share/capstone-prod/capstone/scripts/export_cap_static.py", line 109, in export_cases_by_volume
    export_volume(volume, dest_dir / "redacted")
[...]

PR #2195, opened 2024-02-05
Handle export errors

bensteinberg:
These changes should address #2194.


teovin:
Since we will be gracefully handling the error so the execution continues instead of erroring out and stopping, I think it would be good to log the error inside the except block for informational purposes, future troubleshooting etc.

Also, Jack might have some background on why this error might be being thrown now, I assume the underlying data doesn't change very frequently.

bensteinberg:
Logging would be a good idea, but I'm going to close this PR, as @jcushman is going to look into the underlying problem.

PR #2198, opened 2024-02-06
Fix volume export errors

jcushman:
I moved things around a bit; I think the bug that was happening here is some of our current data has case-to-case citations that should have been filtered out, because the target volumes are out of scope, and that results in the `None` errors we saw in the export. This will filter out citations where the destination case isn't in the export.

I also added record_task_status_for_volume to record errors for any exported volumes in the database.


bensteinberg (approved):
Thanks!

PR #2200, opened 2024-02-07
Fix citation html for cites near page numbers.

jcushman:
I think this avoids the HTML errors on pages like https://cite.case.law/ark/13/409/ where citation html gets mashed up with page number html:

image

This would require running sync_case_body_cache for all volumes, which is a big intervention, so TBD if we want to do that. Possibly there's some way to filter to just cases with html display errors if we can find those.

### DIFF
--- a/capstone/scripts/extract_cites.py
+++ b/capstone/scripts/extract_cites.py
@@ -185,7 +185,15 @@ def extract_citations(case, html, xml):
         # annotate paragraph in html and xml
         if html_annotations:
             for annot_el, annotations in ((xml_els[el_id], xml_annotations), (html_els[el_id], html_annotations)):
-                annot_el.html(annotate(el_text, annotations, annot_el.html(), annotator=annotator))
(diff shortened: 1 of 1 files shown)


bensteinberg (changes requested):
This needs an update to a test fixture, looks like.

mdellabitta:
FWIW, I did some forensics on the HTML of the smaller set of reporters to find cites that linked to other released reporters with Beautiful Soup and it wasn't too bad. Perhaps we could do something similar where we look for the string "href" in the text nodes of each document?

jcushman:
Annoyingly, the tests had detected an actual typo in my PR; I fixed the broken line.

@mdellabitta good idea, it looks like there are about 1000 cases with "href" in them:

```
In [1]: CaseBodyCache.objects.filter(text__contains='href').count()
Out[1]: 907
```

I'm thinking as the next step we could push this change and try updating just those cases and spot check them to see if it worked.

bensteinberg (approved):

PR #2201, opened 2024-02-27
Update cite extraction

jcushman:
This fixes a couple of errors in my previous fix to the html errors in cite extraction -- I tested by running the old and new version on all Mass. cases.

This also makes two substantive changes:
* Sort `data-case-ids`, so cite extraction produces consistent output from run to run.
* Add `data-cite="<corrected citation>"` to each cite link, so you can link from extracted cites to a search page.


bensteinberg (changes requested):
LGTM, but needs some `--recreate_files`

bensteinberg (approved):

PR #2203, opened 2024-02-28
Handle duplicate pages and volumes; add crosscheck function

jcushman:
This updates our export functions to deal with:

- Volumes that have repeating page numbers -- this is fixed by sorting the `0123-01` case ID generation to sort the cases first.
- Multiple volumes in a reporter that have the same volume numbers -- this is fixed by appending `-02`, `-03`.


bensteinberg (approved):
👍

PR #2214, opened 2024-04-05
Update to debian bookworm

jcushman:
This is a little spike on updating old CAP to Debian Bookworm, which would let us run it quite a bit longer internally. I have it down to these remaining test errors:

```
========================================================================================= short test summary info ==========================================================================================
FAILED capapi/tests/test_api.py::test_filter_case_cite_by - assert {31, 32, 33, 34, 35} == {33, 34, 35}
FAILED capapi/tests/test_api.py::test_ngrams_api - KeyError: 'results'
FAILED capdb/tests/test_models.py::test_extract_page_image - AttributeError: 'Document' object has no attribute 'getPageImageList'
FAILED capdb/tests/test_tasks.py::test_export_citation_graph - ModuleNotFoundError: No module named 'scipy'
FAILED cite/tests/test_views.py::test_single_case[pdf] - AttributeError: 'Page' object has no attribute 'getText'
FAILED cite/tests/test_views.py::test_retrieve_page_image - AttributeError: 'Document' object has no attribute 'getPageImageList'
FAILED cite/tests/test_views.py::test_case_editor - subprocess.CalledProcessError: Command '['/usr/local/bin/python', PosixPath('/app/scripts/extract_images.py'), '/app/test_data/downloads/fake_volume.pdf', '1', '3']' returned non-zero exit status 1.
FAILED scripts/tests/test_fastcase.py::test_fastcase_ingest - AssertionError: File /app/test_data/fastcase/A3d/100/100a3d1112_Replace_3.html has changed. Run pytest -k test_fastcase_ingest --recreate_files to update.
FAILED capapi/tests/test_api.py::test_case_detail_pdf - AttributeError: 'Page' object has no attribute 'getText'
FAILED capdb/tests/test_tasks.py::test_pagerank - ModuleNotFoundError: No module named 'scipy'
FAILED capdb/tests/test_build.py::test_pip_compile - subprocess.CalledProcessError: Command '['fab', 'pip-compile']' returned non-zero exit status 1.
```

To get here I turned off a few features that aren't easily supported in python 3.11:

* Disable swagger API specs, never really used, because they depend on coreapi which is unsupported.
* Disable ability to run second-level api queries like cited_by__q, never really used, because they have a small dependency on coreapi and it wasn't immediately obvious how to remove it.
* Disable ngram search in Trends viewer, restricting it to showing API aggregates. I think that's the more transportable part if we adapt that code to CourtListener at some point, and there's no supported python rocksdb client that has the features we used for that.

The upshot is I ran out of time but I think this is doable with a bit more work if we feel like it.


bensteinberg:
This is cool. If the eventual goal is to run the whole system on a bookworm machine, what do you think of upgrading Postgres to the version in bookworm, which is 15? Or is that a separate step?

jcushman:
> This is cool. If the eventual goal is to run the whole system on a bookworm machine, what do you think of upgrading Postgres to the version in bookworm, which is 15? Or is that a separate step?

That would make sense! Yes, the goal is to get a sense of how much additional work would be needed to run on bookworm if we want to -- as opposed to other viable plans like run inside docker for internal use, shutting down in September without the option to spin up later etc. I just spent a couple hours of low-focus time seeing how close I could get, and my sense is like halfway there. For this to be a non-draft we'd want to get the other half, so upgrading the postgres image is a good addition to the todo list.

